PR #1, opened 2020-11-19
Initial zendesk PR

jfrerich:
#### Summary
This is the POC Zendesk cloud app

* README describes the updated functionality
  * zendesk notification when ticket created is posted as a message in Mattermost
  * restapi allows creating a ticket in zendesk and also posts message in Mattermost
* post menu button location added that allows user to create ticket from Mattermost through a modal.
* bot_access_token is stored in local json file (until database designed) #2 

image


src/restapi/responses.ts
```diff
@@ -0,0 +1,71 @@
+import {AppCallResponse, AppSelectOption} from 'mattermost-redux/types/apps';
+
+// createFormResponse returns a form response to create a ticket from a post
```
jfrerich, line 3: Move comment directly above function and match function name

src/restapi/responses.ts
```diff
@@ -0,0 +1,71 @@
...
+class Responses {
+    createForm(message: string): AppCallResponse {
+        const response: AppCallResponse = {
+            type: 'form',
+            form: {
+                title: 'Create Zendesk Ticket',
+                header: 'Create a Zendesk from Mattermost by filling in the required fields and clicking the submit button. Addition text can be added in the `Optional message` form.',
```
jfrerich, line 10: Is it possible to add the zendesk icon to the header?

src/app/store.ts
```diff
@@ -0,0 +1,24 @@
+import fs from 'fs';
+import {jsonStoreFileName} from './constants';
+
+class Store {
+  store: any;
```
jfrerich, line 5: Create custom store type and use instead of 'any'

jfrerich (commented):
Early morning self review

bindings.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "location_id": "/post_menu",
+    "bindings": [
+      {
+        "label": "Create Zendesk Ticket",
+        "description": "Create ticket in zendesk",
+        "icon": "https://raw.githubusercontent.com/jfrerich/mattermost-applet-zendesk/initial-PR/assets/zendesk.svg",
+        "call": {
+          "url": "https://jasonf.ngrok.io/createform",
```
levb, line 10: cc @larkox @mickmister @iomodo 3/5 we switch to "path relative to App's RootURL" everywhere in `Call`, `Notify`, etc. All these URLs exist in the proxied namespace. 0/5 if we keep the field name `url` or change `[url_]path`.

levb: AWS Lambda functions will not be invoked by sending them HTTP requests directly, we'll be using the AWS `Invoke` API, and will map the AppID/path to the specific Lambda in one of our internal accounts.

src/app/app.ts
```diff
@@ -0,0 +1,56 @@
...
+import store from './store';
+
+const username = process.env.ZENDESK_USERNAME as string;
+const token = process.env.ZENDESK_API_TOKEN as string;
+const apiURL = process.env.ZENDESK_URL + '/api/v2' as string;
+
+class App {
+    async createTicketFromPost(ticket, channelId: string, userId: string, postId: string) {
+        const zdClient = zendeskClient(username, token, apiURL);
+        const result = await zdClient.tickets.create(ticket);
```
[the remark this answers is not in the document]
levb: `t`? I am a go guy :)

bindings.json
larkox: I am a bit torn between giving liberty to the app developer or making things simpler and more secure.
If we keep the whole URL, I as an app developer can directly route calls to other services/apps/etc... transparently. If not, we would have to have those endpoints configured and then do a redirect or something similar.
But if the URL is always relative to the root, there are less security concerns since everything is sent "to the same place" (even if then that could potentially get redirected somewhere else). Also, we won't need to even consider having another check of allowed URLs for the app (if a call is trying to access an URL not defined in the manifest, it cannot do it).

The big point here is in my opinion, whether we want to allow inter-app calls and how.

iomodo: At this point we should not allow inter-app calls. That should be done by the OAuth in the future. I'd suggest that lambda functions be called only from the apps-plugin or from the cron job. No one else should be able to call lambdas.

src/utils/util.ts
```diff
@@ -0,0 +1,13 @@
+import fetch from 'node-fetch';
+
+module.exports.openInteractiveDialog = async (dialog) => {
```
larkox, line 3: Why do you need to open an interactive dialog from the api? It should be openable from the call response.

larkox (approved):
A few questions and apps "better practices" (if everything is working as expected, that may be not the case :sweat_smile:)

bindings.json
larkox: What about webhooks from third party services? For example, this app expose webhooks for zendesk to communicate with the app.Will those be also proxied by the apps-plugin? How exactly?

iomodo: We need some kind of proxy for 3rd party webhooks as well. For now apps-plugin will act as one. How exactly TBD.

src/app/app.ts
```diff
@@ -0,0 +1,56 @@
...
+        const zdSubject = values.subject;
+        const mmSignature = '*message created from Mattermost message.*\n';
+
+        const zdMessage = values.additional_message + '\n' +
+            values.post_message + '\n' +
+            mmSignature;
+
+        const ticket: Tickets.CreatePayload = {
+            ticket: {
+                subject: zdSubject,
```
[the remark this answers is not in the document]
jfrerich: done.

sample_webapp_integration.json
```diff
@@ -0,0 +1,12 @@
+{
```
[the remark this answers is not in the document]
jfrerich: done.

src/app/app.ts
```diff
@@ -0,0 +1,56 @@
...
+        const pRes = client.createPost(post);
+    }
+
+    getTicketForPost(values) {
+        const zdSubject = values.subject;
+        const mmSignature = '*message created from Mattermost message.*\n';
+
+        const zdMessage = values.additional_message + '\n' +
+            values.post_message + '\n' +
+            mmSignature;
```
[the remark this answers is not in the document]
jfrerich: Actually, leave mmSignature the same and modify to link back to actual mattermost message post

src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+app.use(cors({ origin: 'http://localhost:8065', credentials: true }));
+
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({
+    extended: true,
+}));
+
+app.get('/', ((req, res) => {
+    res.send('Hello Zendesk');
+}));
```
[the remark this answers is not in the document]
jfrerich: Done.

src/restapi/responses.ts
jfrerich: done.

src/restapi/zd_routes.ts
```diff
@@ -0,0 +1,20 @@
+import express, {Request, Response} from 'express';
+
+import app from '../app/app';
+
+const router = express.Router();
+
+router.get('/', (req: Request, res: Response) => {
+    res.send('ZenDesk!');
+});
```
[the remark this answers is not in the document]
jfrerich: done

src/utils/util.ts
jfrerich: removing. This was leftover from early testing.

src/app/app.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        const post: Post = {
+            message,
+            channel_id: channelId,
+            root_id: postId,
+        };
+
+        const client = mmClient;
+        client.setToken(store.getBotAccessToken());
+        const pRes = client.createPost(post);
```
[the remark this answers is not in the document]
jfrerich: I looked into using the MD response field and it does work, but posts an ephemeral message from system. In this case, I want the Zendesk bot to post a thread response to the root post letting others in Mattermost know that the post was created as a Zendesk ticket.  

I haven't checked extensively to see if this is available, but didn't see anything to accomplish this without using the MM Client4 createPost API .

image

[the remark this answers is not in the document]
larkox: You are right. This usecase should be done with your approach.

Nevertheless, @levb @mickmister should we try to ease this kind of usecase? I have mixed feelings, because the bot may or may not have permission to write in that channel, so it would be "dangerous" to write directly from `plugin-apps` the post.

And on a different note, instead of a System message, should the ephemeral be a message from the app bot?

package.json
```diff
@@ -0,0 +1,49 @@
...
+    "test": "jest --coverage",
+    "rebuild": "rm -rf node_modules; rm package-lock.json; npm i; npm run build"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "body-parser": "^1.19.0",
+    "cors": "^2.8.5",
+    "dotenv": "^8.2.0",
+    "mattermost-redux": "github:mattermost/mattermost-redux#804d4178baeb5f7745da41910bfcf2d4a393c9f0",
```
mickmister, line 19: How close is this to the latest on https://github.com/mattermost/mattermost-redux?

mickmister:
We will need to setup CI on this repo. This includes automated testing, linting, and type checking. @hanzei Is this something trivial with our existing circleci configs?

package.json
```diff
@@ -0,0 +1,49 @@
...
+    "build": "tsc -p .",
+    "build:watch": "tsc -w -p .",
+    "start": "nodemon dist/src/index.js",
+    "test": "jest --coverage",
+    "rebuild": "rm -rf node_modules; rm package-lock.json; npm i; npm run build"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "body-parser": "^1.19.0",
```
mickmister, line 16: Can we use exact package versions to be consistent with other JS repositories we have? Also add an `.npmrc` file like https://github.com/mattermost/mattermost-plugin-starter-template/blob/master/webapp/.npmrc to enforce this.

src/app/app.ts
```diff
@@ -0,0 +1,75 @@
+import {Post} from 'mattermost-redux/types/posts';
+import {AppCall, AppState, AppCallResponse, AppCallValues} from 'mattermost-redux/types/apps';
+import {Tickets, CreatePayload} from 'node-zendesk';
+
+import mmClient from '../mattermost/client';
+import zendeskClient from '../zendesk/client';
```
mickmister, line 6: It'd be awesome to have cross-package imports be able to be absolute imports like this:

```suggestion
import mmClient from 'mattermost/client';
import zendeskClient from 'zendesk/client';
```

This would require changes to `tsconfig.json` for the typescript compiler to know where these are:

```json
{
    "compilerOptions": {
        "paths": {
            "mattermost/*": ["./src/mattermost/*"],
            "zendesk/*": ["./src/zendesk"],
            "src/*": ["./src/*"]
        }
    }
}

File: src/app/app.ts [hunk 16]
```diff
@@ -0,0 +1,75 @@
...
+
+    getTicketForPost(values: AppCallValues): CreatePayload {
+        const mmSignature = '*message created from Mattermost message.*\n';
+
+        const zdMessage = values.additional_message + '\n' +
+            values.post_message + '\n' +
+            mmSignature;
+
+        const ticket: Tickets.CreatePayload = {
+            ticket: {
```
mickmister, line 62: Can we alias this type in a `types/zendesk.ts` file in order to avoid importing `node-zendesk` in `app/app.ts`?

src/utils/util.ts
```diff
@@ -0,0 +1,4 @@
+import fetch from 'node-fetch';
+
+module.exports.makeOption = (name) => ({text: name, value: name});
+module.exports.makeOptions = (names) => names.map((name) => ({text: name, value: name}));
```
mickmister, line 4: `text` will change to `label`

```suggestion
module.exports.makeOption = (name) => ({label: name, value: name});
module.exports.makeOptions = (names) => names.map((name) => ({label: name, value: name}));
```

src/zendesk/client.ts
```diff
@@ -0,0 +1,14 @@
+import zendesk from 'node-zendesk';
+import { Client, ClientOptions } from 'node-zendesk'
+
+function newClient(username: string, token: string, remoteUri: string): Client {
+  const options: ClientOptions = {
+      username: username,
+      token: token,
+      remoteUri: remoteUri,
```
mickmister, line 8:
```suggestion
      username,
      token,
      remoteUri,
```

src/zendesk/client.ts
```diff
@@ -0,0 +1,14 @@
+import zendesk from 'node-zendesk';
+import { Client, ClientOptions } from 'node-zendesk'
```
mickmister, line 2:
```suggestion
import zendesk, {Client, ClientOptions} from 'node-zendesk'
```

src/zendesk/client.ts
```diff
@@ -0,0 +1,14 @@
...
+
+function newClient(username: string, token: string, remoteUri: string): Client {
```
mickmister, line 4:
```suggestion
export default function newClient(username: string, token: string, remoteUri: string): Client {
```

You can also get rid of the export on line 14

tsconfig.json
```diff
@@ -0,0 +1,67 @@
+{
+    "compilerOptions": {
+      /* Basic Options */
+      "target": "ES5", /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
+      "module": "commonjs", /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
```
mickmister, line 5: Can we get rid of the comments and commented out config options?

src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({
+    extended: true,
+}));
+
+app.use('/zendesk', zdRoutes); // Zendesk router
+
+// Mattermost routes must be at root URL level.  The cloud-apps proxy plugin
+// resolves certain paths based on the provided Root URL
+app.use('', mmRoutes);
```
mickmister, line 22: Can we apply secret checks in these routes? Something like:

```js
zdRoutes.use(function (req, res, next) {
    if (zdSecret !== req.headers.Authorization && process.env.NODE_ENV !== 'development') {
        res.status(401).send('Invalid secret');
        return;
    }
    next();
});
```

mickmister, line 22: > Mattermost routes must be at root URL level

Why not mount at `/mattermost` and have the manifest say that `http://localhost:4000/mattermost` is the App's root URL?

src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+app.listen(4000, () => console.log('Listening'));
```
mickmister, line 24:
```suggestion
app.listen(process.env.PORT || 4000, () => console.log('Listening'));
```

src/mattermost/client.ts
```diff
@@ -0,0 +1,8 @@
+import Client from 'mattermost-redux/client/client4.js';
+const client = new Client();
+
+const url = process.env.MM_SITEURL || 'http://localhost:8065';
+client.setUrl(url);
```
mickmister, line 5: We'll definitely want to instantiate clients on the fly, as we will have user tokens and bot tokens to work with. I suggest exporting a function that accepts `(url: string, token: string)`, and returns a new `Client4`.

mickmister (changes requested):
I have about half of the PR reviewed, mostly looks good. Will review more soon.

src/app/app.ts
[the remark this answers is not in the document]
mickmister: Ticket for the last point here https://mattermost.atlassian.net/browse/MM-31083

hanzei:
> We will need to setup CI on this repo. This includes automated testing, linting, and type checking. @hanzei Is this something trivial with our existing circleci configs?

No, our existing circleci config is heavily tied to plugins. I suppose you have to start from scratch. A first step would be to add the targets to the nps scripts. Then creating a circleci config is quite easy.

src/app/calls.ts
```diff
@@ -0,0 +1,71 @@
+import {AppCallResponse, AppSelectOption} from 'mattermost-redux/types/apps';
+
+class Calls {
+    // createForm returns a form response to create a ticket from a post
+    createForm(message: string): AppCallResponse {
```
mickmister, line 5: I'm thinking this should exist in a file `src/forms/create_ticket.ts`

src/app/calls.ts
```diff
@@ -0,0 +1,71 @@
...
+                        type: 'static_select',
+                        options: getStaticSelectOptions(['urgent', 'high', 'normal', 'low']),
+                        is_required: false,
+                        hint: 'priority',
+                        modal_label: 'Priority',
+                    },
+                    {
+                        name: 'additional_message',
+                        type: 'text',
+                        hint: 'Add additional message to the Zendesk ticket',
```
mickmister, line 39: I think this may be better suited to be a description, rather than a hint.

```suggestion
                        description: 'Add additional message to the Zendesk ticket',
```

src/app/calls.ts
```diff
@@ -0,0 +1,71 @@
...
+            form: {
+                title: 'Create Zendesk Ticket',
+                header: 'Create a Zendesk from Mattermost by filling in the required fields and clicking the submit button. Addition text can be added in the `Optional message` form.',
+                footer: 'Message modal form footer',
+                fields: [
+                    {
+                        name: 'subject',
+                        type: 'text',
+                        is_required: true,
+                        hint: 'subject',
```
mickmister, line 17: These hints are a bit useless. Maybe get rid of them?

src/app/calls.ts
```diff
@@ -0,0 +1,71 @@
+import {AppCallResponse, AppSelectOption} from 'mattermost-redux/types/apps';
+
+class Calls {
+    // createForm returns a form response to create a ticket from a post
+    createForm(message: string): AppCallResponse {
+        const call: AppCallResponse = {
+            type: 'form',
+            form: {
+                title: 'Create Zendesk Ticket',
+                header: 'Create a Zendesk from Mattermost by filling in the required fields and clicking the submit button. Addition text can be added in the `Optional message` form.',
```
mickmister, line 10: 0/5

```suggestion
                header: 'Create a Zendesk ticket from Mattermost by filling out and submitting this form. Additional text can be added in the `Optional Message` field.',
```

src/app/calls.ts
```diff
@@ -0,0 +1,71 @@
...
+                        hint: 'subject',
+                        modal_label: 'Subject',
+                    },
+                    {
+                        name: 'type',
+                        type: 'static_select',
+                        options: getStaticSelectOptions(['problem', 'incident', 'question', 'task']),
+                        is_required: false,
+                        hint: 'type',
+                        modal_label: 'Type',
```
mickmister, line 26: Maybe move the `modal_label` to be the second field in each object, rather than one of the last fields

src/app/calls.ts
```diff
@@ -0,0 +1,71 @@
...
+                    },
+                ],
+
+            },
+        };
+        return call;
+    }
+}
+
+function getStaticSelectOptions(values: string[]): AppSelectOption[] {
```
mickmister, line 63: This can be in a utils file

src/app/calls.ts
```diff
@@ -0,0 +1,71 @@
...
+    const options: Array<AppSelectOption> = [];
+    for (const key of values) {
+        options.push({label: key, value: key});
+    }
+    return options;
+}
+
+export default new Calls();
```
mickmister, line 71: Kind of pointless to make this into a `class` that never uses `this`, then exporting instance of the class. Why not just export the `createTicket` function?

tsconfig.json
```diff
@@ -0,0 +1,67 @@
...
+      // "lib": [],                             /* Specify library files to be included in the compilation. */
+      // "allowJs": true,                       /* Allow javascript files to be compiled. */
+      // "checkJs": true,                       /* Report errors in .js files. */
+      // "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
+      "declaration": true, /* Generates corresponding '.d.ts' file. */
```
mickmister, line 10: I don't think this option is being used

src/restapi/mm_routes.ts
```diff
@@ -0,0 +1,49 @@
...
+        const createForm = app.calls.createForm(appCall.context.post.message);
+        res.json(createForm);
+    } else {
+        const message = app.createTicketFromPost(appCall);
+        res.json({});
+    }
+});
+
+router.get('/manifest.json', (req: Request, res: Response) => {
+    res.json(app.manifest);
```
mickmister, line 24: Oh I see the app's fields are being accessed like this, outside of the `App` class. That's a little too implicit imo, as there's no way to know that by looking at the `App` class. Can we have the `App` implement a `getManifest` function to be more explicit?

src/restapi/mm_routes.ts
```diff
@@ -0,0 +1,49 @@
...
+import app from '../app/app';
+
+import {jsonStoreFileName} from '../app/constants';
+
+const router = express.Router();
+
+router.post('/createform', async (req: Request, res: Response) => {
+    const appCall: AppCall = req.body;
+    if (appCall.type === 'form') {
+        const createForm = app.calls.createForm(appCall.context.post.message);
```
mickmister, line 15: `app.calls.createForm` seems like a stretch for a pure function that has no state. Maybe do something like this

```suggestion
        const createTicketForm = newCreateTicketForm(appCall.context.post.message);
```

src/restapi/mm_routes.ts
```diff
@@ -0,0 +1,49 @@
...
+        res.json(createForm);
+    } else {
+        const message = app.createTicketFromPost(appCall);
```
mickmister, line 18: Why assign the return value to this variable and then not use it anywhere? Also we should use `await` here so we can catch an error if it occurs.

src/restapi/mm_routes.ts
```diff
@@ -0,0 +1,49 @@
...
+        res.json({});
```
mickmister, line 19: These API handlers should be wrapped in a `try/catch` block, otherwise the HTTP request would not be closed when an error occurs, as far as I know.

src/restapi/mm_routes.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+});
+
+router.get('/manifest.json', (req: Request, res: Response) => {
+    res.json(app.manifest);
+});
+
+router.get('/bindings', (req: Request, res: Response) => {
+    res.json(app.bindings.getBindings());
```
mickmister, line 28: I think this would read better

```suggestion
    res.json(app.getBindings());
```

src/restapi/mm_routes.ts
```diff
@@ -0,0 +1,49 @@
...
+});
+
+router.post('/install', (req: Request, res: Response) => {
+    // Write json file store
+    fs.writeFile(jsonStoreFileName, JSON.stringify(req.body.values), 'utf8', (err) => {
```
mickmister, line 33: Can this functionality be put into a function in another file, like `storeInstallInfo`?

src/restapi/mm_routes.ts
```diff
@@ -0,0 +1,49 @@
+import fs from 'fs';
+
+import express, {Request, Response} from 'express';
+import {AppCall} from 'mattermost-redux/types/apps';
+
+import app from '../app/app';
+
+import {jsonStoreFileName} from '../app/constants';
```
mickmister, line 8: Can we group the imports like this?

```suggestion
import fs from 'fs';
import express, {Request, Response} from 'express';

import {AppCall} from 'mattermost-redux/types/apps';

import app from '../app/app';
import {jsonStoreFileName} from '../app/constants';
```

src/app/store.ts
```diff
@@ -0,0 +1,35 @@
...
+    store: AppStore;
+
+    constructor() {
+        this.store = {
+            bot_access_token: '',
+            oauth2_client_secret: '',
+        };
+
+        if (fs.existsSync(jsonStoreFileName)) {
+            fs.readFile(jsonStoreFileName, (err, data) => {
```
mickmister, line 20: Seems like we should be saving the store state with this class as well, instead of just having read functionality here. Any store-related `fs` library access should be in this file only 2/5.

Furthermore, we can have an `interface Store` like:

```ts
interface Store {
	getBotAccessToken(): string;
}
```

Then have an implementation like:

```ts
class JSONFileStore implements Store {
	...
}
```

src/app/store.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import {jsonStoreFileName} from './constants';
+
+type AppStore = {
+    bot_access_token: string;
+    oauth2_client_secret: string;
+}
+
+class Store {
+    store: AppStore;
```
mickmister, line 11: Maybe call this `storeData` to avoid confusion?

src/app/bindings.ts
```diff
@@ -0,0 +1,36 @@
...
+    postMenuBindings(): AppBinding {
+        const binding: AppBinding = {
+            location_id: '/post_menu',
+            bindings: [
+                {
+                    label: 'Create Zendesk Ticket',
+                    description: 'Create ticket in zendesk',
+                    icon: 'https://raw.githubusercontent.com/jfrerich/mattermost-applet-zendesk/initial-PR/assets/zendesk.svg',
+                    call: {
+                        url: 'https://jasonf.ngrok.io/createform',
```
mickmister, line 22:
```suggestion
                        url: APP_ROOT_URL + '/createform',
```

src/app/bindings.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    // postMenuBindings returns bindings for the post_menu location
+    postMenuBindings(): AppBinding {
+        const binding: AppBinding = {
+            location_id: '/post_menu',
+            bindings: [
+                {
+                    label: 'Create Zendesk Ticket',
+                    description: 'Create ticket in zendesk',
+                    icon: 'https://raw.githubusercontent.com/jfrerich/mattermost-applet-zendesk/initial-PR/assets/zendesk.svg',
```
mickmister, line 20:
```suggestion
                    icon: 'https://raw.githubusercontent.com/mattermost/mattermost-app-zendesk/initial-PR/assets/zendesk.svg',
```

src/app/bindings.ts
```diff
@@ -0,0 +1,36 @@
...
+                    call: {
+                        url: 'https://jasonf.ngrok.io/createform',
+                        type: 'form',
+                        expand: {
+                            app: '',
```
mickmister, line 25: No reason to include this line

```suggestion
```

src/app/bindings.ts
```diff
@@ -0,0 +1,36 @@
+import {AppBinding, AppState, AppsState} from 'mattermost-redux/types/apps';
+
+class Bindings {
+    // getBindings returns bindings defined for all locations in the app
+    getBindings(): AppState {
```
mickmister, line 5: This seems to make more sense

```suggestion
    getBindings(): AppBinding[] {
```

src/app/bindings.ts
```diff
@@ -0,0 +1,36 @@
...
+        const state: AppsState = [
+            this.postMenuBindings(),
+        ];
+        return state;
+    }
+
+    // postMenuBindings returns bindings for the post_menu location
+    postMenuBindings(): AppBinding {
```
mickmister, line 13: This can be in a flat function `getPostMenuBindings`, as `this` is never used in this method.

src/app/bindings.ts
```diff
@@ -0,0 +1,36 @@
+import {AppBinding, AppState, AppsState} from 'mattermost-redux/types/apps';
+
+class Bindings {
```
mickmister, line 3: This class stores no state in `this`, so we can get rid of the class and use flat functions, unless we expect to provide values in a `constructor`

src/app/http.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const host = process.env.MM_URL || 'http://localhost:8065';
+const pathPrefix = `${host}/plugins/com.mattermost.apps`;
+
+const pathInstall = `${pathPrefix}/install`;
+const pathSubscribe = `${pathPrefix}/subscribe`;
+const pathConnectedInstall = `${pathPrefix}/connected_install`;
+
+// http constains methods for handling requests to and from the
+// mattermost-plugin-apps API
```
mickmister, line 11:
```suggestion
// handles requests to and from the mattermost-plugin-apps API
```

src/app/http.ts
```diff
@@ -0,0 +1,62 @@
...
+        const opts = {
+            method: 'POST',
+            headers: {
+                Authorization: 'Basicjunkencoded',
+                'Content-Type': 'application/json',
+            },
+            body: JSON.stringify('junk'),
+        };
+
+        const r = await fetch(pathInstall, opts);
```
mickmister, line 26: This way the caller can catch an error, or inspect the fetch response

```suggestion
        return fetch(pathInstall, opts);
```

src/app/http.ts
```diff
@@ -0,0 +1,62 @@
...
+        // const res = await fetch(u, {method: 'POST', body: JSON.stringify(dialog)}).then(r => r.json());
+        // return res.json();
+        const opts = {
+            method: 'POST',
+            headers: {
+                Authorization: 'Basicjunkencoded',
+                'Content-Type': 'application/json',
+            },
+            body: JSON.stringify('junk'),
+        };
```
mickmister, line 24: It doesn't seem like any useful stuff is happening here. The commented-out code seems like it would be run elsewhere, and this payload doesn't have any relevant data. I would at least use `JSON.stringify({})`

src/app/http.ts
```diff
@@ -0,0 +1,62 @@
...
+const host = process.env.MM_URL || 'http://localhost:8065';
+const pathPrefix = `${host}/plugins/com.mattermost.apps`;
+
+const pathInstall = `${pathPrefix}/install`;
+const pathSubscribe = `${pathPrefix}/subscribe`;
+const pathConnectedInstall = `${pathPrefix}/connected_install`;
+
+// http constains methods for handling requests to and from the
+// mattermost-plugin-apps API
+class MattermostPluginHTTP {
```
mickmister, line 12: Looks like this file isn't being imported anywhere

README.md
```diff
@@ -1 +1,173 @@
-Start of the zendesk README
+# ZenDesk App
+
+## Quick Start
+
+- clone this repo
+- cp `.env.save` `.env`
```
mickmister, line 6:
```suggestion
- cp `.env.sample` `.env`
```

README.md
```diff
@@ -1 +1,173 @@
...
+- `npm start` - start the node server
+- `npm run build` - builds the dist dir.  (TODO: can this be automated?  Fails if this is not run before `npm start`)
+- `npm run build:watch` (in separate shell) - watch for changing files and report typescript errors
+
+**Current Branches Needed**
+
+The current master branches have not been merged to work with this app and the following are needed until they are mergd with master
+
+mm-webapp: `apps-modals`  
+mm-plugin-apps: `apps-modals`
```
mickmister, line 24:
```suggestion
mattermost-webapp: `feature/cloud-apps`
mattermost-plugin-apps: `master`
```

README.md
```diff
@@ -1 +1,173 @@
...
+
+**Install the app**
+
+`/apps install --url http://localhost:4000/manifest.json --app-secret thisisthesecret`
+
+  After installing the app, a provisioned bot account will be created for user
+  @zendesk and posted in a DM.
```
mickmister, line 31: Any reason why this newline is here?

README.md
```diff
@@ -1 +1,173 @@
...
+
+EndPoint: `http://localhost:4000/createform`
+Method: `post`
+Body (example):
+
+```json
+{
+    "channel_id": "rgiqcxrm8jdjzgj536gb45oh3e",
+    "user_id": "6fiyj9ni9t835dnbni1ddrj93y",
+    "submission": {
```
mickmister, line 48: Should this be updated to use the new Apps form structure?

README.md
```diff
@@ -1 +1,173 @@
...
+1. `Targets tab` > `Add Target`
+1. Select `HTTP` Target
+1. Fill in the following:
+    1. **Title:** Send Mattermost notification when ticket created
+    1. **Url:** `<your_url/zendesk/webhook>`
+    1. **Method:** POST
+    1. **Content Type:** JSON
+
+**Developer Notes:** When testing webhooks locally, you will need to expose your
+localhost:4040 with ngrok
```
mickmister, line 81:
```suggestion
**Developer Notes:** When testing webhooks locally, you will need to expose http://localhost:4000 with ngrok
```

README.md
```diff
@@ -1 +1,173 @@
...
+  team_id?: string;
+  channel_id: string;
+  post_id?: string;
+  root_post_id?: string;
+    props?: {[name: string]: string};
+};
+
+export type AppletNotification = {
+  Subject: string;
+  Context: AppletContext;
```
mickmister, line 131: Need this change in the proxy plugin

```suggestion
  subject: string;
  context: AppletContext;
```

README.md
```diff
@@ -1 +1,173 @@
...
+};
+
+```
+
+## FAQ
+
+### 1. `npm start` fails with warning about rudder in mattermost-redux
```
mickmister, line 138: This is fixed in master now, but probably needs to propagate to mattermost-redux's `feature/cloud-apps` branch

mickmister (changes requested):
Overall functionality looks good. I have lots of requests to improve the code

tsconfig.json
jfrerich: done.

src/app/app.ts
```diff
@@ -0,0 +1,75 @@
...
+class App {
+    bindings: typeof AppState;
+    calls: typeof AppCallResponse;
+    manifest: typeof any;
+
+    constructor() {
+        this.bindings = bindings;
+        this.calls = calls;
+        this.manifest = manifest;
+    }
```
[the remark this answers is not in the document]
jfrerich: yes, it does.  Thanks!

README.md
jfrerich: no.  fixed!

README.md
jfrerich: 👍

src/app/calls.ts
jfrerich: Fair enough.  I was opting for hints because then don't grow the modal vertically.  But I agree for this text it is a description, not a hint :)

src/app/calls.ts
jfrerich: Agree.  They are necessary and are a repeat of the displayed field name

src/app/calls.ts
jfrerich: Much better. Thanks!

src/app/calls.ts
jfrerich: 👍

src/app/bindings.ts
jfrerich: I'll grab this from the manifest.

src/app/bindings.ts
jfrerich: done.

src/app/bindings.ts
jfrerich: done.

src/app/bindings.ts
jfrerich: Agree. Thanks!

src/app/bindings.ts
jfrerich: Agree and flattening this file.

src/app/bindings.ts
jfrerich: Agree here too.  This file should be very simple. It only returns an array of bindings.  Simply bind a location to a call.

src/restapi/mm_routes.ts
jfrerich: I wonder if this function needs to belong to the App class? It returns a list of hardcoded location bindings to a call.  It could almost reside in `src/restapi/` and be imported via `import {getBindings} from './bindings'`.  What do you think?

src/restapi/mm_routes.ts
jfrerich: yes, but lint adds an empty line between the first two.

src/app/http.ts
jfrerich: 🎉 much better wordsmithing

src/app/calls.ts
jfrerich: agree.  I moved it into a utils file inside the app/ folder instead of the utils/ folder.  These will be util functions for the app folder

src/app/calls.ts
jfrerich: similar to bindings.  I had was going to have them all accessible from import `app`.  Then call them via `app.bindings.getBindings()` and `app.calls.createForm()`.  I am going to remove them from the `App` class and have them as standalone files with exported functions.  I just didn't realize how the code would evolve before I implemented it here.

src/restapi/mm_routes.ts
jfrerich: I agree here. Do you think I should also implement method that are used for getting the manifest fields, or is it ok to access them directly?  getManifest().root_url?

src/restapi/mm_routes.ts
jfrerich: Much cleaner. Done.

package.json
jfrerich: done.

src/zendesk/client.ts
jfrerich: done.

src/zendesk/client.ts
jfrerich: done.

src/app/app.ts
jfrerich: I have been working on this for a while with no luck.  Will try a little bit more, but otherwise I'll create a ticket for this item.

jfrerich: Created a ticket for this -> https://github.com/mattermost/mattermost-app-zendesk/issues/7

src/app/calls.ts
jfrerich: Agreed. Done.

tsconfig.json
jfrerich: I'm not 100% sure of the purpose of this option, but verified it does not affect the app. removed.

src/app/http.ts
jfrerich: removed.

src/app/http.ts
jfrerich: This file will be removed

src/app/http.ts
jfrerich: file has been removed.

README.md
jfrerich: I would say no and will be deleting this altogether.  I will replace with an example of the create ticket modal instructions.  This info is left over from when I was first creating the app and using the README as a place to store random information as I was discovering/learning the API.

README.md
jfrerich: Also Removing.  Not needed in the README.

README.md
jfrerich: tested and now running with these branches in my development

src/restapi/mm_routes.ts
jfrerich: bindings has been moved into its own folder `bindings/` similar to the `forms/` folder.

src/app/store.ts
jfrerich: done.

src/restapi/mm_routes.ts
jfrerich: I added this as method in the `app/store.ts`, inside the JSONFileStore class

src/app/store.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/commit/fa8a3e3c7bcebec08dd689fbae3927ceb05f7847

src/restapi/mm_routes.ts
jfrerich: done -> https://github.com/mattermost/mattermost-app-zendesk/commit/fa8a3e3c7bcebec08dd689fbae3927ceb05f7847

src/app/store.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/commit/fa8a3e3c7bcebec08dd689fbae3927ceb05f7847

src/restapi/mm_routes.ts
jfrerich: added the await and also catch any errors within this function call and display in the modal on submit.
https://github.com/mattermost/mattermost-app-zendesk/commit/09732405a9469420730de8af14c8a1ba9d1164e1

src/restapi/mm_routes.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/commit/09732405a9469420730de8af14c8a1ba9d1164e1

src/app/app.ts
```diff
@@ -0,0 +1,75 @@
...
+    async createBotPost(context: AppContext, message: string) {
+        const client = mmClient;
+        client.setToken(store.getBotAccessToken());
+
+        const post: Post = {
+            message,
+            channel_id: context.channel_id,
+            root_id: context.post_id,
+        };
+        const pRes = client.createPost(post);
```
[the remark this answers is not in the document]
jfrerich: Not sure here does the createPost() function throw an error if it cannot complete the request?  I'm not handling the issue and showing anything to the user, but the error should get logged.

src/mattermost/client.ts
jfrerich: yep. Thanks!

src/index.ts
jfrerich: done.

jfrerich: https://github.com/mattermost/mattermost-app-zendesk/commit/be285e89847149fc6810a6312aaefb8d3574a082

src/mattermost/client.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/commit/be285e89847149fc6810a6312aaefb8d3574a082

src/index.ts
jfrerich: Simple solution.  Tested, works, and committed. https://github.com/mattermost/mattermost-app-zendesk/commit/c50f74a0170784181b1c3dc4e421e0a702085bfc

src/utils/util.ts
jfrerich: done.

src/app/app.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    getTicketForPost(values: AppCallValues): CreatePayload {
```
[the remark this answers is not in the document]
jfrerich: types -> in src/types/types.ts -> add CreateTicketFormValues type
mdoel.ts -> in `src/models`/model.ts -> helper constructor functions (newTicketPayloadFromFormSubmission)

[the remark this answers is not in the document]
jfrerich: Please see https://github.com/mattermost/mattermost-app-zendesk/commit/4a43086c1dc657f4533e898dc9744cb1401872c0 for the refactor. I used your suggestion for handling different for types with a base class.  See `forms/handleForms.ts`

package.json
jfrerich: fails using latest master and latest features/cloud branches.

src/app/app.ts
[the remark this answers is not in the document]
jfrerich: I don't need to save the result and now just throw it away.  If an error is thrown, from the createPost() call, the message is shown in the modal

README.md
jfrerich: That is correct.  I'll create a ticket if there isn't one already made.

src/index.ts
jfrerich: I am not 100% sure how to implement this yet.  I see the req coming in for the proxy requests, as `mattermost-app-authorization` with a Bearer JWT token.  I have to figure this out yet.  I'll add to an existing ticket and close this comment for addressing later!

jfrerich: ticket [here](https://mattermost.atlassian.net/browse/MM-31411)

manifest.ts
```diff
@@ -0,0 +1,29 @@
+import {AppsBindings} from 'mattermost-redux/constants/apps';
+
+type RequestedPermissions = 'act_as_user' | 'act_as_bot'
```
jfrerich, line 3: Needs to exist as constants in redux like location constants

manifest.ts
```diff
@@ -0,0 +1,29 @@
...
+type RequestedLocations = AppsBindings.POST_MENU_ITEM | AppsBindings.CHANNEL_HEADER_ICON | AppsBindings.COMMAND | AppsBindings.IN_POST
+
+export type Manifest = {
```
jfrerich, line 6: This manifest type should exist in mm–redux

manifest.ts
```diff
@@ -0,0 +1,29 @@
...
+export function getManifest(): Manifest {
+    const manifest: Manifest = {
+        app_id: 'zendesk',
+        display_name: 'Zendesk',
+        description: 'Zendesk cloud app for Mattermost',
+        root_url: 'https://jasonf.ngrok.io/mattermost',
+        requested_permissions: ['act_as_user', 'act_as_bot'],
+        oauth2_callback_url: 'https://jasonf.ngrok.io/mattermost/oauth2/complete',
+        homepage_url: 'https://github.com/mattermost/mattermost-app-zendesk',
+        requested_locations: ['/channel_header', '/command', '/in_post'],
```
jfrerich, line 26: Replace with AppsBindings constants

manifest.ts
```diff
@@ -0,0 +1,29 @@
...
+    requested_locations: RequestedLocations[];
+}
+
+export function getManifest(): Manifest {
+    const manifest: Manifest = {
+        app_id: 'zendesk',
+        display_name: 'Zendesk',
+        description: 'Zendesk cloud app for Mattermost',
+        root_url: 'https://jasonf.ngrok.io/mattermost',
+        requested_permissions: ['act_as_user', 'act_as_bot'],
```
jfrerich, line 23: Replace with constants after they are created

src/app/app.ts
```diff
@@ -0,0 +1,42 @@
+import {Post} from 'mattermost-redux/types/posts';
+import {AppCall, AppContext} from 'mattermost-redux/types/apps';
+
+import * as mattermost from '../mattermost/client';
```
jfrerich, line 4: Import exact functions instead of *

src/app/app.ts
```diff
@@ -0,0 +1,42 @@
...
+const zdToken = process.env.ZENDESK_API_TOKEN as string;
+const apiURL = process.env.ZENDESK_URL + '/api/v2' as string;
+
+class App {
+    async createTicketFromPost(appCall: AppCall): Promise<string> {
+        const ticket = getTicketForPost(appCall.values);
+
+        const zdClient = zendesk.newClient(username, zdToken, apiURL);
+
+        const result = await zdClient.tickets.create(ticket);
```
jfrerich, line 20: See what type is returned from the client and rename accordingly.   Intention revealing name needed

src/app/app.ts
```diff
@@ -0,0 +1,42 @@
...
+        const user = await zdClient.users.show(result.requester_id);
+
+        const host = process.env.ZENDESK_URL;
+        const message = `${user.name} created ticket [#${result.id}](${host}/agent/tickets/${result.id}) [${result.subject}]`;
+        await this.createBotPost(appCall.context, message);
+    }
+
+    async createBotPost(context: AppContext, message: string) {
+        const url = process.env.MM_SITEURL || 'http://localhost:8065';
```
jfrerich, line 29: Create constants that read the .env file.   Retrieve values from methods not directly from fields

manifest.ts
```diff
@@ -0,0 +1,29 @@
...
+    display_name: string;
+    description: string;
+    root_url: string;
+    requested_permissions: RequestedPermissions[];
+    oauth2_callback_url: string;
+    homepage_url: string;
+    requested_locations: RequestedLocations[];
+}
+
+export function getManifest(): Manifest {
```
jfrerich, line 17: This should be converted to a manifest class which accepts a manifest type.  Constructor reads and verifies fields as required.  Simple helper methods to return fields.

Where would this exist so it could be imported?

src/app/app.ts
```diff
@@ -0,0 +1,42 @@
...
+
+        const result = await zdClient.tickets.create(ticket);
+        const user = await zdClient.users.show(result.requester_id);
+
+        const host = process.env.ZENDESK_URL;
+        const message = `${user.name} created ticket [#${result.id}](${host}/agent/tickets/${result.id}) [${result.subject}]`;
+        await this.createBotPost(appCall.context, message);
+    }
+
+    async createBotPost(context: AppContext, message: string) {
```
jfrerich, line 28: Need an example of use case, but will likely need to post as bot and a user. 

Will need a generic createPost() also.  Propose it accepts user token which will determine who is posting in Mattermost

src/app/model.ts
```diff
@@ -0,0 +1,29 @@
...
+type CreateTicketFormValues = {
+    subject: string;
+    type: string;
+    priority: string;
+    additional_message: string;
+    post_message: string;
+}
+
+export function getTicketForPost(values: CreateTicketFormValues): CreatePayload {
+    const mmSignature = '*message created from Mattermost message.*\n';
```
jfrerich, line 12: Add actual link to mm message

src/bindings/bindings.ts
```diff
@@ -0,0 +1,34 @@
...
+function postMenuBindings(): AppBinding {
+    const binding: AppBinding = {
+        location: AppsBindings.POST_MENU_ITEM,
+        bindings: [
+            {
+                label: 'Create Zendesk Ticket',
+                description: 'Create ticket in zendesk',
+                icon: 'https://raw.githubusercontent.com/mattermost/mattermost-app-zendesk/initial-PR/assets/zendesk.svg',
+                call: {
+                    url: getManifest().root_url + '/createform',
```
jfrerich, line 24: Shoot for manifest.rootURL() using a class or keep and add helper function getManifest().getRootUrL()

src/forms/create_ticket.ts
```diff
@@ -0,0 +1,60 @@
+import {AppCall, AppCallResponse} from 'mattermost-redux/types/apps';
+import {AppFieldTypes} from 'mattermost-redux/constants/apps';
+
+import {makeOptions} from '../utils/utils';
+
+// newCreateTicketForm returns a form response to create a ticket from a post
+export function newCreateTicketForm(call: AppCall): AppCallResponse {
+    const postMessage: string = call.context.post.message;
+    const callResponse: AppCallResponse = {
+        type: 'form',
```
jfrerich, line 10: Replace with constant

src/forms/handleForms.ts
```diff
@@ -0,0 +1,41 @@
...
+
+// BaseForm routes the call type to its corresponding method
+export class BaseForm {
+    constructor(call: AppCall) {
+        this.call = call;
+    }
+
+    handle() {
+        switch (this.call.type) {
+        case 'form':
```
jfrerich, line 13: Replace with constants

src/forms/handleForms.ts
```diff
@@ -0,0 +1,41 @@
...
+    handleForm() {
+        return newCreateTicketForm(this.call);
+    }
+
+    async handleSubmit() {
+        let jsonRes = {};
+        try {
+            await app.createTicketFromPost(this.call);
+        } catch (err) {
+            jsonRes = {
```
jfrerich, line 34: Use TS type and constants for field names

src/forms/handleForms.ts
```diff
@@ -0,0 +1,41 @@
+import app from '../app/app';
+
+import {newCreateTicketForm} from './create_ticket';
```
jfrerich, line 3: Rename file.  handle_forms.ts

jfrerich (commented):
Self review after addressing Michael’s review

manifest.ts
jfrerich: done.

src/forms/handleForms.ts
jfrerich: done.

src/forms/handleForms.ts
jfrerich: done.

src/forms/create_ticket.ts
jfrerich: done.

src/app/app.ts
jfrerich: done.

src/app/app.ts
jfrerich: nope. Will only be posting as the bot.

src/app/app.ts
```diff
@@ -0,0 +1,75 @@
...
+        this.manifest = manifest;
+    }
+
+    async createTicketFromPost(appCall: AppCall): string {
+        const ticket = this.getTicketForPost(appCall.values);
+
+        const zdClient = zendeskClient(username, token, apiURL);
+        const result = await zdClient.tickets.create(ticket);
+        const user = await zdClient.users.show(result.requester_id);
+        const host = process.env.ZENDESK_URL;
```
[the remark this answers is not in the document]
jfrerich: all process constants now accessible from utils/constants.ts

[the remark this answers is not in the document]
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/commit/ff4dc88226cf44eacf3a814792ebcfc4f7e82aec

src/restapi/responses.ts
```diff
@@ -0,0 +1,71 @@
...
+                ],
+
+            },
+        };
+        return response;
+    }
+}
+
+function getStaticSelectOptions(values: string[]): AppSelectOption[] {
+    const options: Array<AppSelectOption> = [];
```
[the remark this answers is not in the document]
jfrerich: done.

src/mattermost/client.ts
```diff
@@ -0,0 +1,8 @@
+import Client from 'mattermost-redux/client/client4.js';
+const client = new Client();
+
+const url = process.env.MM_SITEURL || 'http://localhost:8065';
```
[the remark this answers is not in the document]
jfrerich: done.

src/restapi/mm_routes.ts
```diff
@@ -0,0 +1,58 @@
...
+
+router.post('/createform', (req: Request, res: Response) => {
+    if (req.body.type == 'form') {
+        const createForm = responses.createForm(req.body.context.post.message);
+        res.json(createForm);
+    } else {
+        const {acting_user_id, post_id, team_id, channel_id} = req.body.context;
+        const ticket = app.getTicketForPost(req.body.values);
+        app.createTicketFromPost(ticket, channel_id, acting_user_id, post_id);
+        res.sendStatus(200);
```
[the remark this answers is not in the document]
jfrerich: done.  see forms/create_ticket.ts

jfrerich:
@mickmister, thanks for the thorough review!  I have made quite a few changes since your first PR review.  Hopefully, we can get this merged soon and make tickets for the remaining items, unless there are major issues found.  I would like expand on the functionality by adding other items such as subscriptions and managing all fields when creating a ticket. Thanks again for the review!

src/app/app.ts
jfrerich: Ticket created https://github.com/mattermost/mattermost-app-zendesk/issues/8

README.md
jfrerich: ticket created. https://github.com/mattermost/mattermost-app-zendesk/issues/9

src/app/model.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/issues/10

manifest.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/issues/11#issuecomment-751796413

manifest.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/issues/11#issuecomment-751797277

manifest.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/issues/11#issuecomment-751797277

manifest.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/issues/11#issuecomment-751796413

src/bindings/bindings.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/issues/11

src/store/config.ts
```diff
@@ -0,0 +1,59 @@
...
+        const values = req.body.values;
+        const context: AppContext = req.body.context;
+
+        values.mm_site_url = context.config.site_url;
+
+        fs.writeFileSync(jsonConfigFileStore, JSON.stringify(values), (err) => {
+            if (err) {
+                throw err;
+            }
+        });
```
mickmister, line 47: You can do this to make the function `await`'able, also avoiding use of `writeFileSync`

```suggestion
		return new Promise((resolve, reject) => {
	    	fs.writeFile(jsonConfigFileStore, JSON.stringify(values), (err) => {
				if (err) {
	            	reject(error);
	            	return;
				}
				resolve();
			});
		});	
```

src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({
+    extended: true,
+}));
+
+app.use('/zendesk', zdRoutes); // Zendesk router
+app.use('/mattermost', mmRoutes); // Mattermost router
+
+app.listen(process.env.PORT || 4000, () => console.log('Listening'));
```
mickmister, line 21: If you're interested in making one more small change here

```js
const port = process.env.PORT || 4000;
app.listen(port, () => console.log('Listening on ' + port));
```

src/app/app.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import store from '../store/config';
+
+import {getTicketForPost} from './model';
+
+class App {
+    createTicketFromPost = async (call: AppCall): Promise<string> => {
+        const ticket = getTicketForPost(call.values);
+
+        const zdClient = zendesk.newClient(ENV.zendesk.username, ENV.zendesk.apiToken, ENV.zendesk.apiURL);
```
mickmister, line 16: Can we have the `App` class accept these env vars in a constructor, instead of accessing `utils.ENV` directly? This will make writing tests easier. Anything that relies on the env vars should receive them in a constructor or as a function argument if it is a flat function 2/5.

src/restapi/mm_routes.ts
mickmister: Since this function wants the whole manifest, it needs to have access to the manifest itself. For other fields accessed on the manifest, I think helper methods on `App` would be good like `app.getRootURL()`

src/restapi/mm_routes.ts
mickmister: I'm thinking we should have an uncaught error handler for all routes (in case our try/catches don't catch the error). We can return a call response with a generic error message. Separate ticket for this is good

src/app/bindings.ts
mickmister: We may have different bindings based on if the user is connected to Zendesk for instance. Maybe we do want it in a class if it needs to have user-related information at runtime. Alternatively, it can accept that information as function arguments.

.npmrc
```diff
@@ -0,0 +1,2 @@
+npm_config_save_exact=true
```
mickmister, line 1: I think this should be `save-exact` like the webapp and mobile repos
```suggestion
save-exact=true
```

src/app/app.ts
```diff
@@ -0,0 +1,39 @@
...
+
+        const result = await zdClient.tickets.create(ticket);
+        const user = await zdClient.users.show(result.requester_id);
```
mickmister, line 19: Maybe surround these two calls with `try/catch`. Return an error like `Failed to create ticket` etc. Unless the zendesk client is already returning an error like that

src/app/model.ts
```diff
@@ -0,0 +1,33 @@
...
+    additional_message: string;
+    post_message: string;
+}
+
+export const getTicketForPost = (values: CreateTicketFormValues): CreatePayload => {
+    const mmSignature = '*message created from Mattermost message.*\n' + store.getSiteURL();
+
+    const zdMessage = values.additional_message + '\n' +
+            values.post_message + '\n' +
+            mmSignature;
```
mickmister, line 18: Might want to check if `additional_message` and `post_message` are at least defined here before concatenating strings. Something like

```js
const additionalMessage = values.additional_message || '';
```

This goes for anything else coming from the UI too. We may want some validation here of all fields coming from the `values` object.

src/clients/mattermost.ts
```diff
@@ -0,0 +1,8 @@
+import Client from 'mattermost-redux/client/client4.js';
+
+export const newClient = (token: string, url: string): Client => {
```
mickmister, line 3: Matter of preference, but I think `url` should come first in the arg list here

```suggestion
export const newClient = (url: string, token: string): Client => {
```

src/forms/handle_forms.ts
```diff
@@ -0,0 +1,47 @@
...
+    handle = (): AppCallResponse => {
+        switch (this.call.type) {
+        case 'form':
+            return this.handleForm();
+        case 'lookup':
+            return this.handleLookup();
+        default:
+            return this.handleSubmit();
+        }
+    }
```
mickmister, line 24: Can we have method stubs on base class here like:

```js
handleForm = () => {
    throw new Error('handleForm not implemented');
}
```

src/store/config.ts
```diff
@@ -0,0 +1,59 @@
...
+        };
+
+        if (fs.existsSync(jsonConfigFileStore)) {
+            fs.readFile(jsonConfigFileStore, (err, data) => {
+                if (err) {
+                    throw err;
+                }
+                this.storeData = JSON.parse(data.toString());
+            });
+        }
```
mickmister, line 34: I think we may need to use `readFileSync` here, or have an `async init()` function that reads the json data from the file, so the caller can call that with `await` and have access to the store's data right after the `Store` is instantiated.

src/utils/utils.ts
```diff
@@ -0,0 +1,2 @@
+module.exports.makeOption = (name) => ({label: name, value: name});
+module.exports.makeOptions = (names) => names.map((name) => ({label: name, value: name}));
```
mickmister, line 2: Seems weird to use `export ...` in some cases, and use `module.exports...` here

```suggestion
export const makeOption = (name) => ({label: name, value: name});
export const makeOptions = (names) => names.map(makeOption);
```

mickmister (approved):
@jfrerich Thanks for addressing the PR feedback :+1: I have some non-blocking suggestions

src/utils/utils.ts
jfrerich: Correct.  I made have these changes in another PR but will make them here :)

src/clients/mattermost.ts
jfrerich: done.

.npmrc
jfrerich: done.

src/index.ts
jfrerich: Of course, I am interested in any suggestions. done.

src/store/config.ts
jfrerich: Nice.  I was looking at this for a while and decided to move on after I got my method to work.  thanks

src/app/app.ts
jfrerich: If you don't mind, I will address this in the oauth PR https://github.com/mattermost/mattermost-app-zendesk/pull/12.  I'll also create as a github issue.

src/app/app.ts
jfrerich: I think adding context from the app would be beneficial.  Zendesk is returning an error, it is somewhat cryptic and it would be nice to understand what part of the code failed.  

How about the following prepended message? 
image

jfrerich: Note that I also now catch the error when failing to post a bot post to mattermost and added that to the modal
image

jfrerich: Thanks for this suggestion :)

src/store/config.ts
jfrerich: https://github.com/mattermost/mattermost-app-zendesk/issues/15

PR #16, opened 2021-01-05
[MM-30602] Add Oauth

jfrerich:
#### Summary
This ticket implements oauth 

##### From commit history 
- Branches
  - update to use latest available
  - update README
- README
  - Add how to add Oauth
  - Add available slash commands (connect, disconnect)
- .env.sample
  - only requires Zendesk URL and Client_secrets
- tokens.json
  - storage for user mm (userID) -> zd (access_token) mappings
- Various Typescript Error fixeso
- utils/constants.ts
  - Added constants for zendesk, and app paths
    - both encapsulated with in a routes object for accessing outside
      the package
  - access process.ENV vars from the zendeskENV const
- store/oauth.ts
  - store class for getting/setting/reading/loading user tokens
- Added several index.ts files so all contained variables and functions
and can be exported without reference to the filename

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30602


manifest.ts
```diff
@@ -11,7 +10,7 @@ export type Manifest = {
     requested_permissions: RequestedPermissions[];
     oauth2_callback_url: string;
     homepage_url: string;
-    requested_locations: RequestedLocations[];
+    requested_locations: string[];
```
jfrerich, line 13: We should create the manifest type and also the RequestionLocations type

jfrerich:
opening a clean PR for review

PR #27, opened 2021-02-19
Simplify all call routes to use /

hanzei:
#### Summary
The code didn't take into account that `/mattermost/` was used as a subpath. This PR fixes that issue by removing the subpath.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33053

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -16,6 +16,12 @@
         "/post_menu",
         "/in_post"
     ],
+    "on_install": {
+        "url": "/install"
+    },
+    "bindings": {
+        "url": "/bindings"
+    },
     "functions": [
         {
             "name": "js-function",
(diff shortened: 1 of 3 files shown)


src/manifest.json
```diff
@@ -16,6 +16,12 @@
         "/post_menu",
         "/in_post"
     ],
+    "on_install": {
+        "url": "/install"
+    },
+    "bindings": {
+        "url": "/bindings"
+    },
```
hanzei, line 24: Declaring the path here isn't strictly needed as it's the default path anyway, but I found it easier to understand why a certain path is called. 1/5

jfrerich (approved):
Thanks @hanzei :) I agree the subpath is not necessary and was added in the infancy stages of the app.

levb (approved):

PR #29, opened 2021-02-22
[MM-31508] Rename URL to Path in Call

hanzei:
#### Summary
I've modified `package.json` to use a specified commit in order to test my changes directly. Let me know if I should revert that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31508

### DIFF
--- a/src/bindings/slash_commands.ts
+++ b/src/bindings/slash_commands.ts
@@ -27,7 +27,7 @@ function commandConnect(): AppBinding {
         description: 'Connect your Zendesk account',
         icon: ZDIcon,
         call: {
-            url: Routes.App.BindingPathConnect,
+            path: Routes.App.BindingPathConnect,
         },
     } as AppBinding;
 }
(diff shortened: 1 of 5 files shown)


larkox (approved):

src/bindings/slash_commands.ts
```diff
@@ -27,7 +27,7 @@ function commandConnect(): AppBinding {
         description: 'Connect your Zendesk account',
         icon: ZDIcon,
         call: {
-            url: Routes.App.BindingPathConnect,
+            path: Routes.App.BindingPathConnect,
```
jfrerich, line 30: Looking at the `call` struct here, https://github.com/mattermost/mattermost-plugin-apps/blob/master/apps/call.go#L35-L42 we are still using `url` as the field.  Are we going to change this in master?

hanzei: https://github.com/mattermost/mattermost-plugin-apps/pull/77 will update that.

jfrerich (approved):

PR #30, opened 2021-02-26
Use dist to create bundle

hanzei:
#### Summary
Similar to plugin the command to create the bundle should be behind a `dist` command. I've also added a link the `appsctl`.

#### Ticket Link
Follow up on https://github.com/mattermost/mattermost-app-servicenow/pull/2


jfrerich (approved):
Thanks, @hanzei!

PR #33, opened 2021-03-26
Add 3rd party Oauth support, webhooks, TargetID

jfrerich:
#### Summary

Uses https://github.com/mattermost/mattermost-plugin-apps/pull/114

Uses the proxy app webhook implementation. (https://github.com/mattermost/mattermost-plugin-apps/pull/114)
* Apps can now create the webhook path by adding to the webhook URL extracted from the proxy app with "/{app_id}/{path}"

README
* Update instructions to include new method for creating a Target in Zendesk
* Correct the links for redirect because now pointing to mm-site-url, not the node server hosting the zendesk app  

Add Target Creation Automation
  * Add `/zendesk setup-target` slash command
  * Remove Target ID from config modal
  * create target through the `setup-target` slash command
  * update README
    * remove all the target creation instructions and replace with new
      instructions

`clients/mattermost_proxy_plugin.ts`
* based off of the mmclient in proxy app.  
* used for 
	* KV operations (Set, Get, Delete)
	* Storing Oauth2App
	* Storing Oauth2User token

`bindings`
* configure custom call to include 
  * acting_user to get roles 
  * oauth2_app - to see if the oauth has been configured
  * oauth2_user - to see if the user is connected

Using Oauth2 with latest proxy integration
* client_id & client_secret are stored by the proxy app
* Values are now saved via API and retrieved via expanded oauth2

#### Fixes for Subscriptions (Discovered while testing with Dylan)
* Add instance to the trigger name. The search needs to be specific to the instance so multiple instances can use one ZD account.  

#### Ticket Link
- [x] https://mattermost.atlassian.net/browse/MM-34275 - Use 3rd party Oauth2 support PR
- [x] https://mattermost.atlassian.net/browse/MM-34373 - Remove hardcoded TargetID value from code
- [x] https://mattermost.atlassian.net/browse/MM-33938 - README - add screenshot of ticket create post
- [x] https://mattermost.atlassian.net/browse/MM-33940 - verify and use context.admin_access_token
- [x] https://mattermost.atlassian.net/browse/MM-33448 - Bindings - use context to determine isadmin
- [x] https://mattermost.atlassian.net/browse/MM-33749 - Webhooks - need access to an acting_user_id for creating posts
- [x] https://mattermost.atlassian.net/browse/MM-34721 - Zendesk App: Only sysadmin should be allowed to type Zendesk configure command
[...]

PR #34, opened 2021-03-31
Create clean run of CircleCI (linting fixes)

jfrerich:
#### Summary
Fix Linting errors for a clean CI build


hanzei:
@hanzei Please ping me when this is ready for review

jfrerich:
@hanzei and @mickmister, this PR doesn't solve all of the linting issue, but does resolve a majority of them.  I'd love to get this merged soon so that I can incorporate the changes in https://github.com/mattermost/mattermost-app-zendesk/pull/33

Thanks!

src/forms/subscriptions.ts
```diff
@@ -1,18 +1,21 @@
...
-import {AppSelectOption, AppCall, AppForm, AppField} from 'mattermost-redux/types/apps';
+import {AppSelectOption, AppCallRequest, AppForm, AppField} from 'mattermost-redux/types/apps';
+import {AppContextWithBot} from 'types/apps';
 import {AppFieldTypes} from 'mattermost-redux/constants/apps';
 import Client4 from 'mattermost-redux/client/client4.js';
 
 import {newZDClient, newMMClient, ZDClient} from '../clients';
 import {getStaticURL, Routes} from '../utils';
-import {getBulletedList, makeSubscriptionOptions, makeChannelOptions, getChannelIDFromTriggerTitle, tryPromiseWithMessage} from '../utils/utils';
+import {getBulletedList, makeSubscriptionOptions, makeChannelOptions, getChannelIDFromTriggerTitle, tryPromiseWithMessage, ZDFieldOption} from '../utils/utils';
```
mickmister, line 9: Should `getBulletedList` be named `makeBulletedList`?

src/forms/subscriptions.ts
```diff
@@ -1,18 +1,21 @@
...
+import {ZDTrigger, ZDTriggerCondition} from '../utils/ZDTypes';
 import {SubscriptionFields, ZendeskIcon} from '../utils/constants';
 import {BaseFormFields} from '../utils/base_form_fields';
 
 // newSubscriptionsForm returns a form response to create subscriptions
-export async function newSubscriptionsForm(call: AppCall): Promise<AppForm> {
-    const zdClient = await newZDClient(call.context);
-    const mmClient = newMMClient(call.context).asAdmin();
+export async function newSubscriptionsForm(call: AppCallRequest): Promise<AppForm> {
+    const context: AppContextWithBot = call.context as AppContextWithBot;
```
mickmister, line 16: Either declare the type or use `as`. No need to use both

src/forms/subscriptions.ts
```diff
@@ -362,12 +363,12 @@ class FormFields extends BaseFormFields {
         return this.builder.getFieldLabelByName(SubscriptionFields.SubSelectName);
     }
 
-    getSubTriggerByID(subID: string): any {
-        const triggers = this.getChannelTriggers(this.getSelectedChannelID());
-        return triggers.find((trigger) => String(trigger.id) === subID);
+    getSubTriggerByID(subID: string): ZDTrigger {
+        const triggers: ZDTrigger[] = this.getChannelTriggers(this.getSelectedChannelID());
+        return triggers.find((trigger) => String(trigger.id) === subID) as ZDTrigger;
```
mickmister, line 368: Are we guaranteed to find a trigger with this id here?

src/restapi/fIncomingWebhooks.ts
```diff
@@ -16,30 +18,31 @@ export async function fHandleSubcribeNotification(req: Request, res: Response):
 
     // TODO: we need zendesk bot admin so that admin requests can be made by the
     // bot and not from an actiing user
-    const fakeBotContext: AppContext = {
+    const fakeBotContext: AppContextWithBot = {
+        app_id: getManifest().app_id,
         acting_user_id: 'rgixs6uimp88tq8x8w3yxu3oqe',
-    };
+    } as AppContextWithBot;
```
mickmister, line 24: There's no bot id in this context. Why is it being casted to `AppContextWithBot`?

mickmister (changes requested):
Ideally there would be less `as` going on, but the code does look a bit more defensive now. I have some questions and a few requests

.eslintrc.json
```diff
@@ -63,7 +63,13 @@
...
       2,
       {
-        "properties": "never"
+        "properties": "never",
+        "allow": [
+          "bot_access_token", 
+          "acting_user_id", 
+          "acting_user_access_token",
+          "ticket_field_ids"
+        ]
```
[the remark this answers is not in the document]
hanzei: Does JSON.Parse handle snake_case vs CamelCase correctly? I.e. can we abstract the JSON format away?

src/app/app.ts
```diff
@@ -24,14 +26,14 @@ export interface App {
...
+    context: AppContextWithBot
     values: AppCallValues
 
-    constructor(call: AppCall) {
+    constructor(call: AppCallRequest) {
         this.call = call;
-        this.context = call.context;
-        this.values = call.values;
+        this.context = call.context as AppContextWithBot;
+        this.values = call.values as AppCallValues;
```
hanzei, line 36: Is the `as` really needed?

src/app/app.ts
```diff
@@ -24,14 +26,14 @@ export interface App {
...
+    call: AppCallRequest
+    context: AppContextWithBot
     values: AppCallValues
 
-    constructor(call: AppCall) {
+    constructor(call: AppCallRequest) {
         this.call = call;
-        this.context = call.context;
-        this.values = call.values;
+        this.context = call.context as AppContextWithBot;
```
hanzei, line 35: Not every `Call` has an actin user. How do we deal with that case?

src/app/app.ts
```diff
@@ -120,11 +122,11 @@ class AppImpl implements App {
...
 
         // add bot to team and channel
-        const botUserID = this.context.bot_user_id;
-        const addToTeamReq = adminClient.addToTeam(this.context.team_id, botUserID);
+        const botUserID = this.context.bot_user_id as string;
+        const addToTeamReq = adminClient.addToTeam(this.context.team_id as string, botUserID);
         await tryPromiseWithMessage(addToTeamReq, 'Failed to add bot to team');
 
-        const addToChannelReq = adminClient.addToChannel(botUserID, this.context.channel_id);
+        const addToChannelReq = adminClient.addToChannel(botUserID, this.context.channel_id as string);
```
hanzei, line 129: Why are these `as`'s needed? In which cases are the values `undefined`?

src/app/triggerFromForm.ts
```diff
@@ -9,14 +11,14 @@ interface TriggerFromFrom {
...
-    values: AppFormValues;
+    values: AppCallValues;
     context: AppContext;
-    trigger: any
+    trigger: ZDTrigger
 
-    constructor(call: AppCall) {
-        this.values = call.values as AppFormValues;
+    constructor(call: AppCallRequest) {
+        this.values = call.values as AppCallValues;
```
hanzei, line 19: Why is this needed?

src/app/triggerFromForm.ts
```diff
@@ -9,14 +11,14 @@ interface TriggerFromFrom {
...
         this.context = call.context;
-        this.trigger = {};
+        this.trigger = {} as ZDTrigger;
```
hanzei, line 21: Does it make sense to use `?ZDTrigger` as the type for `trigger` instead?

src/forms/create_ticket.ts
```diff
@@ -64,17 +68,17 @@ class FormFields extends BaseFormFields {
...
     }
 
     async addFormSelectDependentFields(): Promise<void> {
         // get the form id from the selected form field value
-        const formID = this.builder.getFieldValueByName(CreateTicketFields.NameFormsSelect);
+        const formID = this.builder.getFieldValueByName(CreateTicketFields.NameFormsSelect) as unknown;
 
-        const zdFormFieldIDs = this.getTicketFieldIDs(formID);
-        const fieldsListReq = this.zdClient.ticketfields.list();
+        const zdFormFieldIDs = this.getTicketFieldIDs(formID as number);
```
hanzei, line 80: Why not use `ToString` instead of two type assertions?

src/forms/create_ticket.ts
```diff
@@ -154,7 +160,7 @@ class FormFields extends BaseFormFields {
             case ZDFieldTypes.Muliselect: {
                 f.type = AppFieldTypes.STATIC_SELECT;
                 const options = this.isSystemField(field) ? field.system_field_options : field.custom_field_options;
-                f.options = makeOptions(options);
+                f.options = makeOptions(options as ZDFieldOption[]);
```
hanzei, line 163: Why does `makeOptions` not return an `ZDFieldOption[]`?

src/forms/create_ticket.ts
```diff
@@ -171,7 +177,7 @@ class FormFields extends BaseFormFields {
     }
 
     isSystemField(field: Users.Fields.UserField): boolean {
-        return (SystemFields.includes(String(field.type)) || field.system_field_options);
+        return (SystemFields.includes(String(field.type)) || field.system_field_options) as boolean;
```
hanzei, line 180: TS reports:
```
Property 'system_field_options' does not exist on type 'UserField'. Did you mean 'custom_field_options'?ts(2551)
```

src/forms/subscriptions.ts
```diff
@@ -222,17 +223,17 @@ class FormFields extends BaseFormFields {
...
         }
         return [];
     }
 
     // getAnyConditions returns an array of Zendesk ALL trigger conditions
-    getAllConditions(): any[] {
+    getAllConditions(): ZDTriggerCondition[] {
         if (this.getSelectedSubTrigger() && this.getSelectedSubTrigger().conditions) {
-            return this.getSelectedSubTrigger().conditions.all;
+            return this.getSelectedSubTrigger().conditions.all as ZDTriggerCondition[];
```
hanzei, line 236: Why shouldn't the caller deal with the case that `conditions.all` is `undefined´?

src/forms/zendesk_config.ts
```diff
@@ -32,8 +34,8 @@ class FormFields extends BaseFormFields {
     configStore: ConfigStore
     storeValues: AppConfigStore
 
-    constructor(call: AppCall, configStore: ConfigStore, zdClient: ZDClient, mmClient: Client4) {
-        super(call, zdClient, mmClient);
+    constructor(call: AppCallRequest, configStore: ConfigStore, mmClient: Client4) {
+        super(call, {} as ZDClient, mmClient);
```
hanzei, line 38: `BaseFormFields` doesn't require a ´ZDClient`. Can we pass `undefined` here?

src/restapi/fComplete.ts
```diff
@@ -53,11 +53,12 @@ export async function fComplete(req: Request, res: Response): Promise<void> {
...
 // access to the PluginKVStore and retrieves the token and url from state
-function createContextFromState(botToken: string, url: string): AppContext {
-    const context: AppContext = {
+function createContextFromState(botToken: string, url: string): AppContextWithBot {
+    const context: AppContextWithBot = {
+        app_id: '',
         bot_access_token: botToken,
         mattermost_site_url: url,
-    };
+    } as AppContextWithBot;
```
hanzei, line 61: Do we need the `as`?

src/restapi/fConnect.ts
```diff
@@ -1,13 +1,15 @@
...
 import {newOKCallResponseWithMarkdown} from '../utils/call_responses';
 
 import {newConfigStore} from '../store';
 
-import {Routes, createOAuthState, contextFromRequest} from '../utils';
+import {Routes, createOAuthState} from '../utils';
 
 export async function fConnect(req: Request, res: Response): Promise<void> {
-    const context = contextFromRequest(req);
+    const context: AppContextWithBot = req.body.context;
```
hanzei, line 12: 1/5 that I like the helper here.

src/restapi/fDisconnect.ts
```diff
@@ -27,7 +29,7 @@ export async function fDisconnect(req: Request, res: Response): Promise<void> {
 
     // delete the token from the store
     const tokenStore = newTokenStore(context);
-    const deleteTokenReq = tokenStore.deleteToken(context.acting_user_id);
+    const deleteTokenReq = tokenStore.deleteToken(context.acting_user_id as string);
```
hanzei, line 32: The types should make it clear that there is an `actiong_user_id`. Do we need a new type?

src/restapi/fIncomingWebhooks.ts
hanzei: Actually, there should be a bot that we can use, right? See L31.

src/restapi/fIncomingWebhooks.ts
```diff
@@ -1,9 +1,11 @@
 import {Post} from 'mattermost-redux/types/posts';
 
 import {Request, Response} from 'express';
+import {AppContextWithBot} from 'types/apps';
 import {AppContext} from 'mattermost-redux/types/apps';
```
hanzei, line 5: Is this needed any longer?

src/store/oauth.ts
```diff
@@ -31,7 +31,7 @@ class TokenStoreImpl implements TokenStore {
 }
 
 export const newTokenStore = (context: AppContext): TokenStore => {
```
hanzei, line 33: Should this be 
```suggestion
export const newTokenStore = (context: AppContextWithBot): TokenStore => {
```

hanzei (changes requested):

src/app/app.ts
```diff
@@ -120,11 +122,11 @@ class AppImpl implements App {
         const adminClient = newMMClient(this.context).asActingUser();
 
         // add bot to team and channel
-        const botUserID = this.context.bot_user_id;
-        const addToTeamReq = adminClient.addToTeam(this.context.team_id, botUserID);
+        const botUserID = this.context.bot_user_id as string;
```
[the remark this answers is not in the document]
jfrerich: This struct says it is optional

https://github.com/mattermost/mattermost-plugin-apps/blob/master/apps/context.go#L27

src/app/app.ts
jfrerich: @hanzei, the values require the `as` type casting when the type field is optional.  https://github.com/mattermost/mattermost-plugin-apps/blob/master/apps/context.go#L27

jfrerich: @mickmister, maybe the structs tags are not correct and have `omitempty` when that are actually required?

src/restapi/fComplete.ts
```diff
@@ -53,11 +53,12 @@ export async function fComplete(req: Request, res: Response): Promise<void> {
 
 // createContextFromState constructs an AppContext from state. fComplete needs
 // access to the PluginKVStore and retrieves the token and url from state
-function createContextFromState(botToken: string, url: string): AppContext {
-    const context: AppContext = {
+function createContextFromState(botToken: string, url: string): AppContextWithBot {
+    const context: AppContextWithBot = {
```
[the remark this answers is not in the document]
jfrerich: done.

src/restapi/fComplete.ts
jfrerich: Yes. it is needed.  Without, it, the variable `context` will be an object of unknown type

src/store/oauth.ts
jfrerich: yes. thanks and done.

src/forms/subscriptions.ts
jfrerich: done.

src/forms/subscriptions.ts
jfrerich: done.

src/forms/create_ticket.ts
```diff
@@ -35,9 +39,9 @@ export async function newCreateTicketForm(call: AppCall): Promise<AppForm> {
 class FormFields extends BaseFormFields {
     zdTicketForms: ZDFormFieldOption[]
     postMessage: string
-    constructor(call: AppCall, zdClient: ZDClient, mmClient: Client4) {
+    constructor(call: AppCallRequest, zdClient: ZDClient, mmClient: Client4) {
         super(call, zdClient, mmClient);
-        this.postMessage = call.context.post.message;
+        this.postMessage = call?.context?.post?.message as string;
```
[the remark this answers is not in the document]
hmhealey: I'd think the first CallWithContext/CallWithPost part should work, but I can't think of if there's any way to shorthand that. I don't think there's a good generic way to tell TypeScript "yeah, that says optional, but I know it's set" since that sounds like what you're asking about though

src/restapi/fConnect.ts
jfrerich: I don't mind adding this helper func back, but it will need to change.  The AppContext will change depending on the expand values.  I think I'd need to add another input that defined the `type` of context that is expected and what should be returned from the function.

Or.. I could create several helpers that accept a `req`, but return and App Context that is appropriate.  For example 
`contextWithBotFromRequest(req)`
`contextWithPostFromRequest(req)`

jfrerich: By not having the helpers, we can just cast the type when defining the variable.

`const context: AppContextWithBot: = req.body.context`

src/app/app.ts
jfrerich: @hanzei, yes, `values` field in the `CallRequest` struct is an optional field as seen 

https://github.com/mattermost/mattermost-plugin-apps/blob/master/apps/call.go#L55

and 

https://github.com/mattermost/mattermost-redux/blob/master/src/types/apps.ts#L71

src/app/triggerFromForm.ts
jfrerich: Unfortunately. See comment https://github.com/mattermost/mattermost-app-zendesk/pull/34/files#r612100603

I suppose I could create `AppCallRequestWithValues` type that has them as a required field. What do you think?

src/app/triggerFromForm.ts
```diff
@@ -9,14 +11,14 @@ interface TriggerFromFrom {
...
     }
 
     getTrigger(): any {
```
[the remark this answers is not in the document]
jfrerich: Totally makes sense.  It is not defined by the node-zendesk type definitions.  I created a new type and use it now.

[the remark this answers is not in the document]
jfrerich: See commit [`0856c9a` (#34)](https://github.com/mattermost/mattermost-app-zendesk/pull/34/commits/0856c9a66430876490353dec65d824f926f7977e)

src/forms/create_ticket.ts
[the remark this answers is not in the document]
jfrerich: done.

src/forms/create_ticket.ts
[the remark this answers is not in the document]
jfrerich: @mickmister, @hmhealey please see the latest changes to this file and the creation of the following file.

https://github.com/mattermost/mattermost-app-zendesk/blob/lint-fixes/src/types/apps.ts

src/forms/create_ticket.ts
jfrerich: Essentially, this function is converting a `Zendesk` Form Field to an `Apps` `AppField` 

`makeOptions` accepts `ZDFieldOption[]` and returns options for an `AppField`.  Does that make sense?

src/utils/ZDTypes.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export type ZDTriggerCondition = {
+    field: string;
+    operator: string;
+    value?: string;
+}
+
+export type ZDTriggerConditions = {
+    all?: ZDTriggerCondition[];
+    any: ZDTriggerCondition[];
```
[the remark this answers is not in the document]
jfrerich: In the context of how the App currently operates, this type is only used with subscriptions and currently only supports the `any` field.  I added the `all` optional field because it will eventually be supported by the app.

src/app/triggerFromForm.ts
jfrerich: @hanzei, I'm not quite sure I understand your suggestion. I understand the ability to define a field as optional, using the `?` prefix, but I don't know how to use that in the context of defining an empty object for `ZDTrigger`

src/app/app.ts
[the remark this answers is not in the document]
hanzei: It's not optional.

src/app/app.ts
```diff
@@ -120,21 +122,21 @@ class AppImpl implements App {
...
         const botClient = newMMClient(this.context).asBot();
 
-        const post: Post = {
+        const post = {
             message,
             user_id: botUserID,
             channel_id: String(this.context.channel_id),
             root_id: String(this.context.post_id),
-        };
+        } as Post;
```
hanzei, line 139: Is here a good use case for ´Partial`?

src/app/app.ts
hanzei: The `omitempty` is correct. But in all context's that `createBotPost` is called, there will be a `channel_id`.

src/app/triggerFromForm.ts
hanzei: After reading the class a second time I think it should be restructured. There is separate discussion about of the object oriented pattern is really needed, but I'm not going into that.

My suggestion is to the generation of the trigger into the constructor and letting `getTrigger` just returned the class field `trigger`.

src/forms/create_ticket.ts
hanzei: I've misread the code.

Still, I think there is some case handling missing there. If `options` is `undefined` that line with throw an error because `field.system_field_options` might be `undefined`. Can that case happen?

src/forms/create_ticket.ts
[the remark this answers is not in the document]
hmhealey: That works for me.

The other thing I could think to do for that would be to just have two objects passed around: the AppContext and some generic Context that could be modified depending on caller. That simplifies typing, but obviously makes it more annoying to work with since you're passing two objects around instead of once.

src/restapi/fIncomingWebhooks.ts
jfrerich: A refactor has removed the need for this fake context variable

src/app/app.ts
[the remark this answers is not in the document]
jfrerich: fixed.

src/forms/create_ticket.ts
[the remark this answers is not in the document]
jfrerich: To solve this problem, I created some concrete Ctx* types that are used when the request is read.  Any function calls that were using AppContext were changed to access a finite number of inputs. (mattermost_site_url, bot_access_token, .etc)

src/app/app.ts
jfrerich: I created multiple concreate types with required fields and the request in that call function determines which to use.

src/forms/create_ticket.ts
```diff
@@ -64,17 +68,17 @@ class FormFields extends BaseFormFields {
 
     // setState sets state for the current Zendesk Form
     async setState(): Promise<void> {
-        const listReq = this.zdClient.ticketforms.list();
+        const listReq = this.zdClient?.ticketforms.list();
```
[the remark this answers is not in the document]
jfrerich: Yes, you are correct. Thanks.

src/forms/create_ticket.ts
jfrerich: I created a custom ZDField type to handle this case.  https://github.com/mattermost/mattermost-app-zendesk/pull/34/files#diff-317fa1c3e53f7eddff384b174597cc031681277978d4f4ce2e88f13c0834fa5fR29

src/forms/zendesk_config.ts
jfrerich: yes, I had to allow `undefined` in `.eslintrc.json`.   done.

src/restapi/fConnect.ts
jfrerich: I went with creating concrete types that are now used by the call functions.  I could still create the helper functions which would be good for checking the existence of the values.

src/restapi/fDisconnect.ts
jfrerich: The token store has since been removed from this PR when Oauth handling was added to the proxy app.  What this question does bring up is the possible need to expose a proxy plugin endpoint to revoke/delete a token

src/restapi/fIncomingWebhooks.ts
jfrerich: There are multiple new custom context types that are used by the call functions.

.eslintrc.json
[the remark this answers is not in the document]
jfrerich: FYI, I disable this check in one file where it was causing issues (warnings)

src/utils/utils.ts
```diff
@@ -22,17 +23,20 @@ type ZDSubscriptionFieldOption = {
...
 // trigger title
 export const getChannelIDFromTriggerTitle = (title: string): string => {
     const re = new RegExp(SubscriptionFields.RegexTriggerTitle);
-    const match = title.match(re);
+    const match = title.match(re) || '';
     if (!match) {
-        console.log('malformed Mattermost Trigger title', match[1]);
+        console.log('malformed Mattermost Trigger title', title);
     }
     return match[1];
```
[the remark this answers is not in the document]
jfrerich: @mickmister, is it appropriate to throw and error instead so the function exists?

```js
throw new Error('Failed to get user access_token' + title);
```

src/utils/utils.ts
```diff
@@ -22,17 +23,20 @@ type ZDSubscriptionFieldOption = {
 
 const getDisplaySubTitleOption = (option: ZDSubscriptionFieldOption): string => {
     const re = new RegExp(SubscriptionFields.RegexTriggerTitle);
-    const newTitle = option.title.match(re)[2];
-    return newTitle;
+    const newTitle = option.title.match(re) || '';
+    if (!newTitle) {
+        console.log('malformed Mattermost Trigger title', newTitle);
+    }
+    return newTitle[2];
```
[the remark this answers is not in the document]
jfrerich: converted to throw an error

src/forms/subscriptions.ts
jfrerich: converted to throw an error

src/app/triggerFromForm.ts
jfrerich: done.

src/app/app.ts
jfrerich: This code has been removed. The bot can post to any channel now and does not need to be added to a channel or team

src/forms/create_ticket.ts
jfrerich: remove the `as unknown` and now casting formID as a string.

src/forms/subscriptions.ts
jfrerich: done.

src/forms/create_ticket.ts
jfrerich: You cannot save a drop-down field in Zendesk without at least one option.  I'm not sure the best way to handle the check for that case.  If options are undefined, I suppose that I could just not show that field?

jfrerich: Ok. I added a check for defined options and if they don't exist, the field is added, but without any options. Again, this case should never occur because of the zendesk restrictions when created a drop box.

src/forms/create_ticket.ts
```diff
@@ -170,14 +200,14 @@ class FormFields extends BaseFormFields {
         });
     }
 
-    isSystemField(field: Users.Fields.UserField): boolean {
-        return (SystemFields.includes(String(field.type)) || field.system_field_options);
+    isSystemField(field: ZDUserField): boolean {
+        return (SystemFields.includes(String(field.type)) || Boolean(field.system_field_options));
```
mickmister, line 204: What's the need to cast via `String()` here?

src/forms/subscriptions.ts
```diff
@@ -83,18 +97,20 @@ class FormFields extends BaseFormFields {
...
+            results.push(this.addChannelTrigger(trigger));
         }
+        await Promise.all(results);
     }
 
-    async addChannelTrigger(trigger: any): Promise<void> {
+    async addChannelTrigger(trigger: ZDTrigger): Promise<void> {
         const parsedTitle = parseTriggerTitle(trigger.title);
         const channelID = parsedTitle.channelID;
         const channel = await this.mmClient.getChannel(channelID);
```
mickmister, line 116: Could we do some caching here during the request, so we don't fetch the same channel more than once?

PR #35, opened 2021-04-05
fix fetch error

jfrerich:
#### Summary
Fix fetch error. node does not include fetch by default.  Add it to fix the FAQ issue

#### Ticket Link
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,13 +95,13 @@ We need to create the Zendesk HTTP target which will send webhook trigger notifi
     1. **Method:** POST
     1. **Content Type:** JSON
 1. Test that the target is valid
-    1. Select `Test target` in the pulldown
+    1. Select `Test target` in the dropdown
     1. Click `Submit` button
     1. Leave JSON body in the floating window empty
     1. Click `Submit` button in floating window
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -7,6 +7,8 @@ import bodyParser from 'body-parser';
 // import zdRoutes from './restapi/zd_routes';
 import mmRoutes from './restapi/mm_routes';
 
+global.fetch = require('node-fetch');
```
mickmister, line 10: I recommend putting this import above any local imports (like `./restapi/mm_routes`), since it is changing the environment that modules in that tree may be using.

README.md
```diff
@@ -112,33 +112,19 @@ To provision this PR to AWS run `make dist` to generate the App bundle and then
...
 Oauth2 service needs to be turned on in `config/config.json`
 Through system console -> enable oauth2 service provider
 
-
 ```json
 "EnableOAuthServiceProvider": true,
 ```
 
-### 3. Need a branched version of the node-zendesk client so that we can query
+### 2. Need a branched version of the node-zendesk client so that we can query
```
mickmister, line 126: Can you fork the repo so there's no patch needed in the local copy?

PR #38, opened 2021-04-06
remove outdated FAQ

jfrerich:
#### Summary
- move global.fetch before all local imports
- remove FAQ for branched node-zendesk.  This was already done.
	- https://github.com/mattermost/mattermost-app-zendesk/blob/45ab240937cb797025a71ce43861ef3605e1e0d5/package.json#L23

#### Ticket Link
N/A


jfrerich:
@mickmister, this PR is for your comments in https://github.com/mattermost/mattermost-app-zendesk/pull/35!  Should be a quick review :)

src/index.ts
```diff
@@ -4,11 +4,11 @@ import express from 'express';
 import serverless from 'serverless-http';
```
mickmister, line 4: Can we move the `dotenv.config()` line to just above the `global.fetch` line? It seems odd to have a function call above other imports. You are supposed to have the `dotenv.config()` [as early as possible](https://github.com/motdotla/dotenv#usage), but really that just means before any of *your* code. So putting it after external imports is fine.

mickmister (changes requested):
Changes look great :+1: though I have an unrelated request

src/index.ts
jfrerich: Actually, I think we can remove the import of `dotenv` and the `dotenv.config()` because we no longer read from a `.env` file.

mickmister (approved):

PR #44, opened 2021-04-27
Fix multiple ticket issues in Zendesk

jfrerich:
#### Summary
This PR fixes another round to Jira Ticket issues

#### Ticket Link
- [x] https://mattermost.atlassian.net/browse/MM-35210 - MM Apps - Zendesk show root ID error when creating a ticket from a reply thread
- [x] https://mattermost.atlassian.net/browse/MM-35209 - Access token in config is invalidated
- [x] https://mattermost.atlassian.net/browse/MM-35259 - MM APPS - Zendesk Caution the user who is authorizing subscriptions against disconnecting
- [x] https://mattermost.atlassian.net/browse/MM-35258 - MM Apps - Make IDs in Zendesk human readable
- [x] https://mattermost.atlassian.net/browse/MM-35273 - MM Apps - Zendesk after install and Configuration - Only help and connect where showing
- [x] https://mattermost.atlassian.net/browse/MM-35299 - MM Apps - Zendesk remove "Brand" from subscription event list
- [x] https://mattermost.atlassian.net/browse/MM-35314 - Subscriptions - make group_id human readable
- [x] https://mattermost.atlassian.net/browse/MM-35321 - Subscriptions - remove system_post_all role and add bot to team and channel through mm client
- [x] https://mattermost.atlassian.net/browse/MM-35336 - MM Apps - Zendesk there should be user feedback that the user has connected to Zendesk

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -142,10 +142,7 @@ class AppImpl implements App {
         const actingUserClient = newMMClient(this.mmOptions).asActingUser();
         const userID = this.context.acting_user_id;
 
-        let rootID = this.context.post.root_id;
-        if (this.context.post.root_id === '') {
-            rootID = this.context.post.id;
-        }
+        const rootID = this.context.post.root_id || this.context.post.id;
         const post = {
             message,
             user_id: userID,
@@ -160,10 +157,7 @@ class AppImpl implements App {
     createBotPost = async (message: string): Promise<void> => {
         const botUserID = this.context.bot_user_id;
 
-        let rootID = this.context.post.root_id;
-        if (this.context.post.root_id === '') {
-            rootID = this.context.post.id;
-        }
+        const rootID = this.context.post.root_id || this.context.post.id;
         const post = {
             message,
             user_id: botUserID,
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -123,6 +123,16 @@ class AppImpl implements App {
             actionType = 'create';
         }
 
+        const adminClient = newMMClient(this.mmOptions).asAdmin();
+
+        // add bot to team and channel
+        const botUserID = this.context.bot_user_id;
+        const addToTeamReq = adminClient.addToTeam(this.context.team_id, botUserID);
+        await tryPromiseWithMessage(addToTeamReq, 'Failed to add bot to team');
+
+        const addToChannelReq = adminClient.addToChannel(botUserID, this.context.channel_id);
+        await tryPromiseWithMessage(addToChannelReq, 'Failed to add bot to channel');
+
         // Any zendesk error will produce an error in the modal
         let msg: string;
         try {
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -1,4 +1,5 @@
 import {Post} from 'mattermost-redux/types/posts';
+import {Channel} from 'mattermost-redux/types/channels';
 import {AppCallValues, AppCallResponse, AppCallRequest} from 'mattermost-redux/types/apps';
 
 import {
@@ -23,6 +24,7 @@ import {newTriggerFromForm} from './triggerFromForm';
 export interface App {
     createTicketFromPost(): Promise<AppCallResponse>;
     createZDSubscription(): Promise<AppCallResponse>;
+    createBotDMPost(message: string): Promise<void>;
 }
 
 class AppImpl implements App {
@@ -178,6 +180,24 @@ class AppImpl implements App {
         const botClient = newMMClient(this.mmOptions).asBot();
         const createPostReq = botClient.createPost(post);
         await tryPromiseWithMessage(createPostReq, 'Failed to create bot post');
+    }
+
+    createBotDMPost = async (message: string): Promise<void> => {
+        const botUserID = this.context.bot_user_id;
+        const actingUserID = this.context.acting_user_id;
+
+        const botClient = newMMClient(this.mmOptions).asBot();
+        const channel: Channel = await botClient.createDirectChannel([botUserID, actingUserID]);
+
+        const post = {
+            message,
+            user_id: botUserID,
+            channel_id: channel.id,
+            root_id: '',
+        } as Post;
+
+        const createPostReq = botClient.createPost(post);
+        await tryPromiseWithMessage(createPostReq, 'Failed to create bot DM post');
     }
 
     hasFieldErrors(errors: FieldValidationErrors): boolean {
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -142,29 +142,38 @@ class AppImpl implements App {
         const actingUserClient = newMMClient(this.mmOptions).asActingUser();
         const userID = this.context.acting_user_id;
 
+        let rootID = this.context.post.root_id;
+        if (this.context.post.root_id === '') {
+            rootID = this.context.post.id;
+        }
         const post = {
             message,
             user_id: userID,
             channel_id: String(this.context.channel_id),
-            root_id: String(this.context.post_id),
+            root_id: String(rootID),
         } as Post;
 
         const createPostReq = actingUserClient.createPost(post);
-        await tryPromiseWithMessage(createPostReq, 'Failed to create post');
+        await tryPromiseWithMessage(createPostReq, 'Failed to create acting user post');
     }
 
     createBotPost = async (message: string): Promise<void> => {
         const botUserID = this.context.bot_user_id;
+
+        let rootID = this.context.post.root_id;
+        if (this.context.post.root_id === '') {
+            rootID = this.context.post.id;
+        }
         const post = {
             message,
             user_id: botUserID,
             channel_id: String(this.context.channel_id),
-            root_id: String(this.context.post_id),
+            root_id: String(rootID),
         } as Post;
 
         const botClient = newMMClient(this.mmOptions).asBot();
         const createPostReq = botClient.createPost(post);
-        await tryPromiseWithMessage(createPostReq, 'Failed to create post');
+        await tryPromiseWithMessage(createPostReq, 'Failed to create bot post');
     }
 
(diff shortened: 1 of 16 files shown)


mickmister:
There are some linting errors happening:

```
src/app/app.ts:145:22 - error TS2532: Object is possibly 'undefined'.

145         let rootID = this.context.post.root_id;
```

I recommend using `this.context.root_id || this.context.post_id`

jfrerich:
@mickmister, thanks for the early responses. I am implementing a couple more fixes and was going to do a final review.  Very much appreciate the early feedback

jfrerich:
@catalintomai, @mickmister there are obvious areas of improvements to DRY up the code.  I implementing working code, but did not have time to continue further.  I just wanted y'all to be aware before the review :)

src/app/app.ts
```diff
@@ -123,6 +123,16 @@ class AppImpl implements App {
...
 
+        const adminClient = newMMClient(this.mmOptions).asAdmin();
+
+        // add bot to team and channel
+        const botUserID = this.context.bot_user_id;
+        const addToTeamReq = adminClient.addToTeam(this.context.team_id, botUserID);
+        await tryPromiseWithMessage(addToTeamReq, 'Failed to add bot to team');
+
+        const addToChannelReq = adminClient.addToChannel(botUserID, this.context.channel_id);
+        await tryPromiseWithMessage(addToChannelReq, 'Failed to add bot to channel');
```
hanzei, line 134: This seems like a hack to me. An app should be contained by it's own channel membership and not add itself to all channels, even though it can do that. Requiring `post_all_as_bot` is more honest.

src/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
     "app_id": "com.mattermost.zendesk",
     "app_type": "aws_lambda",
-    "version": "0.2.0",
+    "version": "0.2.4",
```
hanzei, line 4: Revert?

src/restapi/fDisconnect.ts
```diff
@@ -14,6 +15,16 @@ export async function fDisconnect(req: Request, res: Response): Promise<void> {
         botAccessToken: context.bot_access_token,
         mattermostSiteUrl: context.mattermost_site_url,
     };
+
+    // get the saved service account config zendesk access_token
+    const config = await newConfigStore(context.bot_access_token, context.mattermost_site_url).getValues();
+    const configOauthToken = config.zd_oauth_access_token;
+    const text = 'This mattermost account is connected via oauth2 to Zendesk for subscription functionality.  The account cannot be disconnected until the access token in the configuration is updated to a new user access token.';
```
hanzei, line 22:
```suggestion
    const text = 'This mattermost account is connected via oauth2 to Zendesk for subscription functionality. The account cannot be disconnected until the access token in the configuration is updated to a new user access token.';
```

src/utils/utils.ts
```diff
@@ -138,7 +138,7 @@ export function getCheckBoxesFromTriggerDefinition(definitions: any): checkBox[]
         // restrict possible checkbox values for simplicity
         // - custom checkbox has only two possible values, but not supported by 'change'
         // - group === requester also not easily determined
-        if (action.values && action.group === 'ticket' && action.subject !== 'follower' && !isCustomField) {
+        if (action.values && action.group === 'ticket' && action.subject !== 'follower' && !isCustomField && action.subject !== 'brand_id') {
```
mickmister, line 141: Would it help to make these strings into constants?

src/restapi/fIncomingWebhooks.ts
```diff
@@ -55,22 +53,147 @@ export async function fHandleSubcribeNotification(req: Request, res: Response):
...
+        }
+        mappedArray.push(event);
+    }
+
+    // resolve all the Promises
+    await Promise.all(promiseEvents).then((mappedEvents) => {
+        for (const mappedEvent of mappedEvents) {
+            mappedArray.push(mappedEvent);
+        }
+    });
```
mickmister, line 112: How many requests can this end up being here? Also, is it possible that `mappedEvent` is null?

Is it possible to grab multiple types of values with one request, to limit the number of requests? Some way of reusing data, to avoid over-fetching?

src/restapi/fIncomingWebhooks.ts
```diff
@@ -55,22 +53,147 @@ export async function fHandleSubcribeNotification(req: Request, res: Response):
...
+    requests.push(tryPromiseWithMessage(currReq, 'Failed to get current Zendesk user'));
+
+    if (event.previous_value) {
+        const prevReq = zdClient.users.show(event.previous_value);
+        requests.push(tryPromiseWithMessage(prevReq, 'Failed to get previous Zendesk user'));
     }
+
+    await Promise.all(requests).then((values) => {
+        event.value = values[0].name;
+        if (values.length === 2) {
```
mickmister, line 166: Lots of duplicated code in these functions. This may or may not make things simpler:

```js
type ZDEvent = {value: any, previous_value?: any};

function processEventValue(zdClientMethod: (value: any) => any, event: ZDEvent): ZDEvent {
    const requests: any[] = [];
    const currReq = zdClientMethod(event.value);
    
    // actually don't know how to compute the correct error message though. Probably needs to also be passed in
    requests.push(tryPromiseWithMessage(currReq, 'Failed to fetch current ticket form'));

    if (event.previous_value) {
        const prevReq = zdClientMethod(event.previous_value);
        requests.push(tryPromiseWithMessage(prevReq, 'Failed to fetch previous ticket form'));
    }

    await Promise.all(requests).then((values) => {
        event.value = values[0].name;
        if (values.length === 2) {
            event.previous_value = values[1].name;
        }
    });
    return event;
}
```

Then in `mapIDsToTextValues`:

```js
if (event.field_name === 'ticket_form_id') {
    promiseEvents.push(processEventValue(zdClient.ticketforms.show, event));
    continue;
}
```

Seems like the code may be more difficult to read this way though. And if anything needs to change on any specific method (may already be the case), then you wouldn't be able to reuse this across the board.

mickmister (changes requested):
Mostly just some questions, and some ambitious suggestions. Let me know what you think 🙂

catalintomai (approved):

src/app/app.ts
hanzei: > Arguably, if the acting user doesn't have permission to add users to the current channel, the shouldn't have access to create subscriptions in the current channel. 

That the important point to me.

src/restapi/fDisconnect.ts
hanzei: My comment was just about the double white space not more.

Wtr i18n: Either the i18n files are served via static assets or just places next to the code. I'm a bit concerned about i18n in AWS as loading the files take a lot of time and memory. I suppose it's fine for small apps. Most importantly, only the needed language should be loaded, not all of them.

jfrerich: Double space removed.

src/types/apps.ts
```diff
@@ -80,6 +80,7 @@ export type ExpandedActingUser = AppContext & {
 
 export type CtxExpandedBotAdminActingUserOauth2User = ExpandedActingUser & ExpandedOauth2User & ExpandedBot & ExpandedAdmin
 export type CtxExpandedBotAdminActingUserOauth2UserChannel = ExpandedActingUser & ExpandedOauth2User & ExpandedBot & ExpandedAdmin & ExpandedChannel
+export type CtxExpandedBotAdminActingUserOauth2UserChannelPost = ExpandedActingUser & ExpandedOauth2User & ExpandedBot & ExpandedAdmin & ExpandedChannel & ExpandedPost
```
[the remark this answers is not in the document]
jfrerich: Ticket created for this issue.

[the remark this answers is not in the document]
jfrerich: https://mattermost.atlassian.net/browse/MM-35363

src/restapi/fIncomingWebhooks.ts
jfrerich: ticket created https://mattermost.atlassian.net/browse/MM-35364

src/manifest.json
jfrerich: done.

src/restapi/fIncomingWebhooks.ts
jfrerich: I don't have better solution

src/app/app.ts
mickmister: @hanzei The solution I mentioned uses the acting user's token to do the add the bot to the channel. So if the acting user does not have permissions to add the bot to the channel, then the subscription add would/should fail. No non-permissive channel adding going on here, as long as we use the acting user token to do the add.

jfrerich: I have changed the code to:

1. check that the acting user is a MM system admin
2. the MM client now uses the acting_user_id, not the general admin access_token

src/utils/utils.ts
jfrerich: This will get revamped in v2 of the plugin.  Ticket created so this doesn't get lost
https://mattermost.atlassian.net/browse/MM-35365

src/restapi/fIncomingWebhooks.ts
jfrerich: To better answer your question.  

The max number of requests would be on a changed event with all of the field names.. A request is needed for the current name and the previous, so a max number of requests is ..

* 4 * 2 = 8 requests.  
  * 4 (number of fields than need mapped from id to human readable)
  * 2 (retrieve the current and previous values)

**Is it possible that `mappedEvent` is null?**
Each of the `mappedEvents` in the `promiseEvents` array are already await and resolved.  Also each request in the getXXXName() function is checked for an error.  

If any request fails, the webhook will not send and the failed request message is logged.

DHaussermann (approved):
Tested and passed 
- All issues resolved
- No further functional testing needed
LGTM!

PR #45, opened 2021-04-28
await storeConfigInfo calls

mickmister:
#### Summary

This PR makes it so any calls to the `storeConfigInfo` function are `await`'d


iomodo (approved):

jfrerich (approved):
Awesome! I discussed the issue with Catalin and how this seems to affect AWS Lambda when a function is kicked off, but not awaited.

PR #47, opened 2021-05-01
Attempt to fix get bindings flakiness

catalintomai:


jfrerich (approved):
Thanks @catalintomai.  I like the way that calling `storeConfigInfo` is always `awaited`.  I think a future enhancement of the `ppClient` methods should also await all of their calls without putting the requirement on the caller.

catalintomai:
Yeah, I have to educate myself more on what the best patterns are in these cases, but my belief is that we want to await closer to the source not at the top.

src/store/config.ts
```diff
@@ -24,8 +24,8 @@ class ConfigStoreImpl implements ConfigStore {
         this.storeData = {} as AppConfigStore;
     }
 
-    storeConfigInfo(store: AppConfigStore): Promise<void> {
-        return this.ppClient.kvSet(StoreKeys.config, store);
+    async storeConfigInfo(store: AppConfigStore): Promise<void> {
+        await this.ppClient.kvSet(StoreKeys.config, store);
```
mickmister, line 28: Returning and awaiting will do the same thing here, it's just that returning is cleaner imo. Since `kvSet` is already returning a promise, returning its value will also return a promise, not requiring this function to be declared as `async`. It's a matter of preference though since we are returning `Promise<void>`.

src/restapi/fConfig.ts
```diff
@@ -39,7 +39,7 @@ export async function fSubmitOrUpdateZendeskConfigSubmit(req: Request, res: Resp
         const targetID = cValues.zd_target_id;
         const storeValues = call.values as AppConfigStore;
         storeValues.zd_target_id = targetID;
-        await configStore.storeConfigInfo(storeValues);
+        configStore.storeConfigInfo(storeValues);
```
mickmister, line 42: Why is this no longer awaited? That will cause issues with lambda that were solved with https://github.com/mattermost/mattermost-app-zendesk/pull/45. The lambda function will not wait for that http request to be finished because we aren't waiting for the promise to resolve here.

src/restapi/fTarget.ts
```diff
@@ -59,7 +59,7 @@ async function updateOrCreateTarget(zdClient: ZDClient, context: CtxExpandedBotA
     cValues.zd_target_id = zdTarget.id;
 
     // save the targetID
-    await config.storeConfigInfo(cValues);
+    config.storeConfigInfo(cValues);
```
mickmister, line 62: Same question here

mickmister (commented):
I think we are re-introducing some bugs here by not `await`ing some calls to `storeConfigInfo`

mickmister:
> but my current belief is that we want to await closer to the source not at the top

@catalintomai The await needs to "bubble up" to the top either way. This can be through explicit `async/await` expressions throughout the call stack, or by returning a chain of promises (which is what is implicitly happening with `async/await`). So `await`s (or returning promises from other functions) should be happening at both the source and the top of the call stack. Otherwise the lambda function will cut those promise chains off before they finish/resolve.

src/store/config.ts
catalintomai: I can see both ways, since `storeConfigInfo ` is basically a wrapper around the `kvset` one could indeed return the promise and let the upper-level calls await on it - or, instead - await all the way to the top (with the function that returns the promise at the bottom of the chain - the only one).

src/restapi/fConfig.ts
catalintomai: @mickmister  - I think you are right. We'd need to await this as well, since otherwise the execution could continue and we run the risk that the lambda invoker returns before the offloaded `storeConfigInfo` wrapper executes. Though - it it prob. a smaller risk than before (when the `kvset` post was not awaited at all) - and that is prob. why we haven't seen regressions on the issues originally fixed by your  original PR - it is still a risk. @jfrerich, feel free to re-add the await prefix where I removed it.

src/restapi/fTarget.ts
catalintomai: see above.

PR #49, opened 2021-05-02
Remove channel picker

jfrerich:
#### Summary
This PR removes the channel picker from the subscriptions modal. There is some flakiness and the complexity will be removed for v1 of the zendesk app.

* remove channel picker select field
* when client searches for triggers, include the channelID so that the client results are scoped only to the current channel
* when a new sub is selected:
  * show hint `Name of your subscription` instead of actual text 

#### Ticket Link
- [x] https://mattermost.atlassian.net/browse/MM-35378 - MM Apps - Znedesk remove channel picker in subscriptions modal
- [x] https://mattermost.atlassian.net/browse/MM-35379 - MM Apps - Remove default text from the subscription name field for new subscriptions form

PR #51, opened 2021-05-03
Store is_agent in the oauth2user store 

jfrerich:
#### Summary
This PR limits the bindings if a user is not a Zendesk Agent. 

##### Configure Requirements
* Must be MM syadmin

##### Create Ticket Requirements
* Must be ZD `admin` or `agent`
* Does not require MM Sysadmin

##### Subscriptions
* Must be ZD `admin`
* Does not require MM Sysadmin

##### Setup-Target
* Must be ZD `admin`
* Must be MM `System Admin` 
  * This step can be run multiple times, but shouldn't be.  Dylan and I decided to restrict to MM System Admins the slash command does not show for most agents.

##### Disconnect
* Any Zendesk user can disconnect
* If the saved service access_token is the one of the user trying to disconnect, the user receives the following message:
  * `This mattermost account is connected via oauth2 to Zendesk for subscription functionality and cannot be disconnected until the access token is updated to a new user access token. Please have another connected Mattermost System Admin user with Zendesk Admin privileges run /zendesk setup-target to update the access_token`;

##### Connect
* Any Zendesk user can connect
* Zendesk user role is stored in the oauth2 user store so bindings can be determined by role, but without the need for a Zendesk client request on every bindings update
* All connected users get a DM with `You have successfully connected your Zendesk account!`;
  *  user is not ZD `admin` or `agent` they also get this additional message `This app currently supports Zendesk admin and agent accounts and does not provide any features for end-user accounts`;

The store has been modified to add a `role` field and the token data is now stored in the `token` field.

I did not want to modify the return data from Zendesk by adding the `role` to the actual returned data.  This is why I modified the store structure to be the following:

```ts
{
   'token': <zd-token-data>,
   'role': <admin | agent | end-user>
}
```

#### The shared access token was previously saved via the following steps:
- user configures through modal
- user connects
- user runs `/zendesk me` command to get their `access_token`
  - **this step visibly shows the access token and is a security risk**
- user copies value, opens the config modal, pastes and clicks save.

#### This PR updates the method by adding the access_token when the user runs the `/zendesk setup-target` command
* updates README 
* removes the `/zendesk me` command from bindings, but left the `fMe.ts` file for future use.
* Updates message to user that tries `/zendesk disconnect`, but is the user with the shared `access_token`. 
  * include instructions to have another user run `/zendesk setup-target` to transfer the `access_token`.

#### Ticket Link
- [ ] https://mattermost.atlassian.net/browse/MM-33376


jfrerich:
@levb, @mickmister this PR is needed for the release today.  

@dylan, this PR is based off https://github.com/mattermost/mattermost-app-zendesk/pull/49 which also removed the channel picker.

src/restapi/fMe.ts
```diff
@@ -3,7 +3,7 @@ import {Request, Response} from 'express';
 import {newOKCallResponseWithMarkdown} from '../utils/call_responses';
 
 export async function fMe(req: Request, res: Response): Promise<void> {
-    const message = '`access_token:` ' + req.body.context.oauth2.user.access_token;
+    const message = '`access_token:` ' + req.body.context.oauth2.user.token.access_token;
```
levb, line 6: we should NOT display the full value of the token, maybe the last 4 characters? (separate issue/PR)

levb (approved):

src/restapi/fMe.ts
jfrerich: This was added so that an admin can get their token and store it in the config.  It's the only use case this is needed.  A better idea might be to have a `/zendesk storeConfigToken` that would:
1. get the current user access_token
2. verify they are an admin
3. store the token in the config

jfrerich: @levb, @aaronrothschild, @DHaussermann what are your thoughts on fixing this for `p1`?  Is this a breaking change that prevents releasing until it is fixed?

jfrerich: It is a pretty trivial fix. I'll push a PR soon regardless 👍🏼

jfrerich: PR created for this issue https://github.com/mattermost/mattermost-app-zendesk/pull/52

src/utils/ZDTypes.ts
```diff
@@ -55,3 +56,8 @@ export type ZDTokensResponse = {
     previious_page: string
     count: number
 }
+
+export type StoredOauthUserToken = {
+    token: ClientOAuth2.Data,
+    is_agent: boolean
+}
```
mickmister: There's some inconsistency in this file for using commas, semicolons, or using no delimiter. 1/5 to pick one to be consistent

src/utils/constants.ts
```diff
@@ -190,6 +190,11 @@ ZDFieldValidation[ZDFieldTypes.Regex] = {
     RegexError: 'Regex field not valid',
 };
 
+export const ZDRoles = {
+    admin: 'admin',
+    agent: 'agent',
+};
```
mickmister, line 196: Maybe use an enum here

```suggestion
export enum ZDRoles {
    admin = 'admin',
    agent = 'agent',
};
```

src/restapi/fMe.ts
jfrerich: I've pushed a commit to this PR that changes this functionality. Previously, the `/zendesk me` command existed for the sole purpose of exposing a user's ZD `access_token` in an ephemeral post.  One user would then copy that token value and paste it in the configuration modal, for the service ZD `access_token`.  This was a one time setup for webhooks, but the `/zendesk me` command was available for all users.  @levb's concern was that the full value was on display, which was true and a bad implementation.  I have since pushed a commit to this PR that changes the way the service account access_token is saved. 

A user now only need to run the `/zendesk setup-target` command which will:
1. setup the target in Zendesk
2. save their access_token as the service account access token

jfrerich: This was the PR that added the new capability.  It was merged into this PR.  https://github.com/mattermost/mattermost-app-zendesk/pull/52

src/utils/ZDTypes.ts
jfrerich: changed all to remove commas and semicolons.

DHaussermann (approved):
Tested and passed
- `me` command removed
- Creating target now automatically makes a target without the user seeing the value and manually storing
- No regression to the exiting mechanism that prevents the service account from disconnecting
- Target can be transferred to a new service account
- Any ZD "Agent" create tickets regardless of MM Role 
- Any ZD "Admin" can create a Trigger regardless of MM Role 
- A User who is **both** a ZD Admin and  MM Admin can create a target 
- Some regression testing of subscriptions targeting. No issues found 

LGTM!

src/forms/create_ticket.ts
```diff
@@ -22,7 +22,7 @@ const omitFields = ['Group', 'Status'];
 export async function newCreateTicketForm(call: AppCallRequest): Promise<AppForm> {
     const context = call.context as CtxExpandedBotAdminActingUserOauth2User;
     const zdOptions: ZDClientOptions = {
-        oauth2UserAccessToken: context.oauth2.user.access_token,
+        oauth2UserAccessToken: context.oauth2.user.token.access_token,
```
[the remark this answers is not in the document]
jfrerich: It's not possible because the user cannot open the create modal unless they are connected with checks for the token

src/utils/constants.ts
jfrerich: I have a ticket for using enums, but want to spend time understanding the benefits and tradeoffs first

jfrerich: added into https://mattermost.atlassian.net/browse/MM-33764

src/restapi/fMe.ts
mickmister: So why is that step needed though? We already have the value so why do we need the user to copy it from us, and paste it into another form (or command) to us?

src/forms/create_ticket.ts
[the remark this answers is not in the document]
mickmister: I don't think we should have cross-endpoint assumptions though. This can be an arbitrary http request coming to the app. 2/5 we should at least have a ticket to make sure each endpoint's assumptions are verified by that same endpoint.

src/utils/constants.ts
mickmister: The only tradeoff I can think of is that it wouldn't be usable in a javascript file, but since the whole project is typescript we are good on that front.

mickmister (approved):

src/forms/create_ticket.ts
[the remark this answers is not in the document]
jfrerich: ticket created https://mattermost.atlassian.net/browse/MM-35440

PR #52, opened 2021-05-03
Allow updating shared access_token via slash command and not through config modal

jfrerich:
#### Summary

This PR udpates the method for adding the shared `access_token` needed for subscriptions

#### The shared access token was previously saved via the following steps:
- user configures through modal
- user connects
- user runs `/zendesk me` command to get their `access_token`
  - **this step visibly shows the access token and is a security risk**
- user copies value, opens the config modal, pastes and clicks save.

#### This PR updates the method by adding the access_token when the user runs the `/zendesk setup-target` command
* updates README 
* removes the `/zendesk me` command from bindings, but left the `fMe.ts` file for future use.
* Updates message to user that tries `/zendesk disconnect`, but is the user with the shared `access_token`. 
  * include instructions to have another user run `/zendesk setup-target` to transfer the `access_token`.

#### Ticket Link
No ticket.  Originated from this comment.
https://github.com/mattermost/mattermost-app-zendesk/pull/51#discussion_r625255370

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,8 @@ This step requires a Zendesk-connected Mattermost system administrator and uses 
 
 Note, the access token is only used to read ticket information when a subcription is triggered. This token will not post on behalf of the user.
 
-1. `/zendesk setup-target` - This command will set up a Zendesk target pointing to your Mattermost instance (it only needs to be run once). The command also uses saves the access_token for the acting user for subscriptions functionality.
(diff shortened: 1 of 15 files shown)


mickmister:
> removes the `/zendesk me` command from bindings, but left the `fMe.ts` file for future use.

The `/zendesk me` command (or equivalent) is required for GDPR compliance. We need to show the user everything we are storing about them (without showing sensitive credentials), as well as deleting all of this information when they disconnect.

DHaussermann:
Tested as part of https://github.com/mattermost/mattermost-app-zendesk/pull/51

PR #53, opened 2021-05-03
Add lint rule for avoiding floating promises

mickmister:
#### Summary

This PR adds an eslint rule that checks for unhandled promises.

After enabling the rule, I ran `npm run lint` to find any existing issues with promises. There were three issues it pointed out. One was for a function that was inherently not async that was being declared as async, and not awaited when called. The other two were un-awaited calls to `storeConfigInfo`.

@catalintomai I ran the lint rule on the commit that preceded https://github.com/mattermost/mattermost-app-zendesk/pull/47, and it pointed out the issues that were fixed in that PR. Namely, the calls to `ppClient.kvSet` that were previously unawaited would be caught by this lint rule, which your PR fixed those issues. The lint rule should catch any issues that may cause lambda issues going forward.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -509,6 +509,7 @@
       "files": ["**/*.tsx", "**/*.ts", "**/*.js"],
       "parser": "@typescript-eslint/parser",
       "rules": {
+        "@typescript-eslint/no-floating-promises": ["error"],
         "@typescript-eslint/no-var-requires": 0,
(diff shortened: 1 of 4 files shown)


hanzei (approved):
LGTM :+1:

catalintomai (approved):

jfrerich (approved):
Needs a conflict resolved, but this is great!  Thanks

PR #65, opened 2021-07-09
[MM-33762] Use markdown instead of posting in text textarea box

jfrerich:
#### Summary
This PR makes use of field-level Markdown. The message used a textbox to show the message, but now uses markdown which allows formatting and hyperlinking the trigger link.

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33762

### DIFF
--- a/src/forms/subscriptions.ts
+++ b/src/forms/subscriptions.ts
@@ -113,13 +113,16 @@ class FormFields extends BaseFormFields {
     async getTriggers(): Promise<any> {
         // modified node-zendesk to allow hitting triggers/search api
         // returns all triggers for all channels and teams
-        let search = SubscriptionFields.PrefixTriggersTitle;
-        search += SubscriptionFields.RegexTriggerInstance;
-        search += this.call.context.mattermost_site_url;
-        search += SubscriptionFields.RegexTriggerTeamID;
-        search += this.call.context.team_id;
-        search += SubscriptionFields.RegexTriggerChannelID;
-        search += this.call.context.channel_id;
+        const search = [
+            SubscriptionFields.PrefixTriggersTitle,
+            SubscriptionFields.RegexTriggerInstance,
+            this.call.context.mattermost_site_url,
+            SubscriptionFields.RegexTriggerTeamID,
+            this.call.context.team_id,
+            SubscriptionFields.RegexTriggerChannelID,
+            this.call.context.channel_id,
+        ].join('');
+
         const client = this.zdClient as ZDClient;
         const searchReq = client.triggers.search(search) || '';
         return tryPromiseWithMessage(searchReq, 'Failed to fetch triggers');
@@ -274,10 +277,13 @@ class FormFields extends BaseFormFields {
     // addErrorMessageField adds a text field with a message when a trigger has
     // conditions not supported by the app
     addErrorMessageField(link: string): void {
-        let md = 'The following condition fields are not currently supported by the app. Please visit the trigger link to modify the conditions for this subscription\n';
-        md += makeBulletedList('Unsupported Fields', this.unsupportedFields) + '\n';
-        md += makeBulletedList('Unsupported Field Operators', this.unsupportedOperators) + '\n';
-        md += '##### Trigger Link\n' + link;
+        const md = [
+            'The following condition fields are not currently supported by the app. Please visit the trigger link to modify the conditions for this subscription',
+            makeBulletedList('Unsupported Fields', this.unsupportedFields),
+            makeBulletedList('Unsupported Field Operators', this.unsupportedOperators),
+            '##### Trigger Link',
+            link,
+        ].join('\n');
 
         const f: AppField = {
             name: SubscriptionFields.UnsupportedFieldsTextName,
--- a/src/forms/subscriptions.ts
+++ b/src/forms/subscriptions.ts
@@ -241,7 +241,7 @@ class FormFields extends BaseFormFields {
         if (this.unsupportedFields.length !== 0 || this.unsupportedOperators.length !== 0) {
             const host = this.zdHost;
             const trigger = this.getSelectedSubTrigger();
-            const url = `${host}agent/admin/triggers/${trigger.id}`;
+            const url = `${host}/agent/admin/triggers/${trigger.id}`;
             this.addErrorMessageField(url);
             return false;
         }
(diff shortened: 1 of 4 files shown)


hanzei (approved):
Nice :+1:

src/forms/subscriptions.ts
```diff
@@ -274,20 +274,15 @@ class FormFields extends BaseFormFields {
...
-        let text = 'The following condition fields are not currently supported by the app. Please visit the trigger link to modify the conditions for this subscription';
-        text += '\n\n';
-        text += makeBulletedList('Unsupported Fields', this.unsupportedFields);
-        text += '\n\n';
-        text += makeBulletedList('Unsupported Field Operators', this.unsupportedOperators);
-        text += '\n\n' + link;
+        let md = 'The following condition fields are not currently supported by the app. Please visit the trigger link to modify the conditions for this subscription\n';
+        md += makeBulletedList('Unsupported Fields', this.unsupportedFields) + '\n';
+        md += makeBulletedList('Unsupported Field Operators', this.unsupportedOperators) + '\n';
+        md += '##### Trigger Link\n' + link;
```
mickmister, line 280: I'm curious what you think about this strategy of using an array and `.join('\n')` when we're creating multiline markdown like this

```suggestion
        const md = [
            'The following condition fields are not currently supported by the app. Please visit the trigger link to modify the conditions for this subscription',
            makeBulletedList('Unsupported Fields', this.unsupportedFields),
            makeBulletedList('Unsupported Field Operators', this.unsupportedOperators),
            '##### Trigger Link',
            link,
        ].join('\n');
```

src/utils/utils.ts
```diff
@@ -96,7 +96,7 @@ export function baseUrlFromContext(mattermostSiteUrl: string): string {
 export function makeBulletedList(pretext: string, items: string[]): string {
     let text = '* ' + items.join('\n* ');
     if (pretext) {
-        text = `###  ${pretext}\n` + text;
+        text = `####  ${pretext}\n` + text;
```
mickmister, line 99: Not sure if this is suggestion is easier to read. Just seems weird that we're using backtick syntax and "plus" string concatenation.

```suggestion
        text = `####  ${pretext}\n${text}`
```

src/forms/subscriptions.ts
jfrerich: Yes, I like that and used this `<array>.join()` format in other places of the code base. Thanks for the suggestion.

src/utils/utils.ts
jfrerich: done.

src/forms/subscriptions.ts
```diff
@@ -274,20 +277,18 @@ class FormFields extends BaseFormFields {
...
-        text += '\n\n';
-        text += makeBulletedList('Unsupported Field Operators', this.unsupportedOperators);
-        text += '\n\n' + link;
+        const md = [
+            'The following condition fields are not currently supported by the app. Please visit the trigger link to modify the conditions for this subscription',
+            makeBulletedList('Unsupported Fields', this.unsupportedFields),
+            makeBulletedList('Unsupported Field Operators', this.unsupportedOperators),
+            '##### Trigger Link',
+            link,
+        ].join('\n');
```
larkox, line 286: You were using before double line breaks, instead of just one. Not sure if this change is intended.

src/restapi/fHelp.ts
```diff
@@ -8,9 +8,11 @@ import {CommandTrigger} from '../utils/constants';
...
 
 export async function fHelp(req: Request, res: Response): Promise<void> {
-    let helpText = getHeader();
-    helpText += getCommands(req.body);
-    helpText += getPostText();
+    const helpText = [
+        getHeader(),
+        getCommands(req.body),
+        getPostText(),
+    ].join('\n');
```
larkox, line 15: Apparently, you were not joining these string with line breaks. Not sure if this change is intended.

larkox (approved):

src/forms/subscriptions.ts
jfrerich: The text that was used before was very generic and just thrown into a text box. This new style is fine :)

src/restapi/fHelp.ts
jfrerich: Nice catch.  The line breaks are not needed here.

src/restapi/fHelp.ts
```diff
@@ -29,35 +31,38 @@ function getCommands(call: AppCallRequest): string {
...
 
 function getPostText(): string {
-    let text = h5('Post Menu Options');
-    text += 'click the (...) on a post\n';
-    text += addBullet('Create Zendesk Ticket');
-    text += addBullet('Zendesk Subscriptions');
-    return text;
+    return [
+        h5('Post Menu Options'),
+        'click the (...) on a post\n',
```
mickmister, line 54: Maybe make a `text` function that appends the newline so it's more consistent with the other things in this list. Also, should "click" be capitalized?

src/forms/subscriptions.ts
mickmister: It may even be a good idea to create a `markdown` function like:

```ts
export const joinMarkdown = (...lines: string[]) => {
    return lines.join('\n');
}
```

This would allow you to create multi-line markdown like this:

```ts
const md = joinMarkdown(
    'The following condition fields are not currently supported by the app. Please visit the trigger link to modify the conditions for this subscription',
    makeBulletedList('Unsupported Fields', this.unsupportedFields),
    makeBulletedList('Unsupported Field Operators', this.unsupportedOperators),
    '##### Trigger Link',
    link,
);
```

Not sure if that improves the code with that name though. There's probably a better name than `joinMarkdown`

mickmister (approved):
LGTM :+1:

mickmister:
Using the functionality from https://github.com/mattermost/mattermost-webapp/pull/8219, we can put this markdown in the `error` property for the specific form field. I wonder how that would look, maybe the header size text can be made smaller to utilize the error message space/styling.

src/forms/subscriptions.ts
jfrerich: I did add the two small functions. `textLine` and `JoinLines`.  I'm sure there are MD libraries that would also help alleviate these steps. :)

src/restapi/fHelp.ts
jfrerich: done.

jfrerich:
Jira issue created for https://github.com/mattermost/mattermost-app-zendesk/pull/65#issuecomment-879601582

https://mattermost.atlassian.net/browse/MM-37166

PR #67, opened 2021-07-09
remove trailing slash if one is added to the end of the zd url

jfrerich:
#### Summary
When a user adds the ZD URL in the configuration modal remove the trailing `/` if one exists.  If the ZD link is of the form `http://mm.zendesk.com//agent/tickets/123` (note the double slash) the link will not resolve correctly when clicked.

This PR removes a trailing slash before saving the configuration values to the KV store.

The PR also checks the Zendesk URL is valid and returns an error if the link is not
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36946

### DIFF
--- a/src/restapi/fConfig.ts
+++ b/src/restapi/fConfig.ts
@@ -35,6 +35,7 @@ export async function fSubmitOrUpdateZendeskConfigSubmit(req: Request, res: Resp
         const targetID = cValues.zd_target_id;
         const zdOauth2AccessToken = cValues.zd_oauth_access_token;
         const storeValues = call.values as AppConfigStore;
+        storeValues.zd_url = storeValues.zd_url.replace(/\/$/, '');
         storeValues.zd_target_id = targetID;
         storeValues.zd_oauth_access_token = zdOauth2AccessToken;
         await configStore.storeConfigInfo(storeValues);
--- a/src/restapi/fConfig.ts
+++ b/src/restapi/fConfig.ts
@@ -36,7 +36,7 @@ export async function fSubmitOrUpdateZendeskConfigSubmit(req: Request, res: Resp
         const targetID = cValues.zd_target_id;
         const zdOauth2AccessToken = cValues.zd_oauth_access_token;
         const storeValues = call.values as AppConfigStore;
-        storeValues.zd_url = storeValues.zd_url.replace(/\/+$/, '');
+        storeValues.zd_url = storeValues.zd_url.replace(/\/$|(?<!\/)\/+$/, '');
 
         try {
             await verifyUrl(storeValues.zd_url);
--- a/src/restapi/fConfig.ts
+++ b/src/restapi/fConfig.ts
@@ -5,8 +5,9 @@ import {AppCallRequestWithValues, CtxExpandedBotActingUserAccessToken} from '../
 import {newConfigStore, AppConfigStore} from '../store/config';
 import {newAppsClient} from '../clients';
 import {newZendeskConfigForm} from '../forms';
-import {newOKCallResponseWithMarkdown, newFormCallResponse, newErrorCallResponseWithMessage} from '../utils/call_responses';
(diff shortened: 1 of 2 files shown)


catalintomai (approved):

src/restapi/fConfig.ts
```diff
@@ -35,6 +35,7 @@ export async function fSubmitOrUpdateZendeskConfigSubmit(req: Request, res: Resp
         const targetID = cValues.zd_target_id;
         const zdOauth2AccessToken = cValues.zd_oauth_access_token;
         const storeValues = call.values as AppConfigStore;
+        storeValues.zd_url = storeValues.zd_url.replace(/\/$/, '');
```
mickmister, line 38: What validation do we have that verifies that this is a valid zendesk URL? Can we ping zendesk's API to verify that the URL is valid?

mickmister (approved):
LGTM :+1:

src/restapi/fConfig.ts
mickmister: Note that if there is more than one trailing slash, it will only remove the last one.

jfrerich: I updated the replace statement to remove all ending `/`'s

jfrerich: I haven't added any validation for the provided URL.  What if I just ping the login `https://mattermost1616716488.zendesk.com/access/unauthenticated` when responds with status `200` if the site exists?

mickmister: Yeah that sounds good

jfrerich: image

@mickmister, I added a check for a valid URL.  This is pretty nice because the modal hyperlinks the URL in the error :)

src/restapi/fConfig.ts
```diff
@@ -35,6 +36,16 @@ export async function fSubmitOrUpdateZendeskConfigSubmit(req: Request, res: Resp
         const targetID = cValues.zd_target_id;
         const zdOauth2AccessToken = cValues.zd_oauth_access_token;
         const storeValues = call.values as AppConfigStore;
+        storeValues.zd_url = storeValues.zd_url.replace(/\/$|(?<!\/)\/+$/, '');
```
mickmister, line 39: Maybe we should have a comment saying exactly what the goal of the regex is. I have to say that it's gotten a little more complicated and I wouldn't know what it's doing upon inspection

src/restapi/fConfig.ts
```diff
@@ -35,6 +36,16 @@ export async function fSubmitOrUpdateZendeskConfigSubmit(req: Request, res: Resp
...
+
+        try {
+            await verifyUrl(storeValues.zd_url);
+        } catch (error) {
+            callResponse = newErrorCallResponseWithFieldErrors({zd_url: error.message});
```
mickmister, line 44: Nice 🙂

src/restapi/fConfig.ts
```diff
@@ -43,3 +54,15 @@ export async function fSubmitOrUpdateZendeskConfigSubmit(req: Request, res: Resp
     }
     res.json(callResponse);
 }
+
+const verifyUrl = async (url: string) => {
+    const verifyURL = url + Routes.ZD.AccessURI;
+    try {
+        const resp = await fetch(verifyURL, {method: 'post'});
```
mickmister, line 61: Since the method is `POST`, showing the (clickable) URL in the error message doesn't necessarily help because clicking the URL will have the user send a `GET` request to zendesk, in case they wanted to test it manually. 0/5 on whether that matters

mickmister (approved):
I have some non-blocking questions but LGTM!

jfrerich:
removed QA label.  Dylan will test at release

src/restapi/fConfig.ts
jfrerich: done.

src/restapi/fConfig.ts
jfrerich: @mickmister I added a pretty length comment with a link to the example. This was actually a pretty useful catch by codeQL

PR #68, opened 2021-07-21
[MM-35339] Subscription names should be unique per channel

jfrerich:
#### Summary
This PR adds validation that a subscription name is unique before Updating or Creating a new subscription. 

The subs options for a channel are passed as state so that a zendesk API call does not need to be called more than once, when the modal is initially opened.

The validation message displays an error when the subscription name is already being used and must be changed by the user.

image

#### Testing

I tested the following:

##### 1. user had previously saved subscriptions with the same name (ex. `subA` and `subA`)
* save some subscriptions that have the same name  (before using this PR, because this PR prevents this from happening)
  * (ex. `subA` and `subA`)
* using this PR branch, select one of the `subA` subscriptions from the dropdown. 
* when the save button is clicked, the message will appear, informing the user they need to edit the name
* change the name and save
* verify subscription has been updated and the name changed.

Note: this is a case that might occur if someone has already created subscriptions with the same name. The subscriptions will work as is but the user will be asked to rename a sub if this case occurs and while trying to update.

##### 2. user creates a new subscription
* no change - subscription created

##### 3. user tries changing the name of a subscription to another existing subscription
* setup - subA and subB exist
* subA is selected in the dropdown
* subA name changed to subB and click save
* field error should show

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35339

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -1,6 +1,6 @@
 import {Post} from 'mattermost-redux/types/posts';
 import {Channel} from 'mattermost-redux/types/channels';
-import {AppCallValues, AppCallResponse, AppCallRequest} from 'mattermost-redux/types/apps';
+import {AppCallValues, AppCallResponse, AppCallRequest, AppSelectOption} from 'mattermost-redux/types/apps';
 
 import {
     newErrorCallResponseWithFieldErrors,
@@ -8,7 +8,7 @@ import {
     newOKCallResponseWithMarkdown,
     newErrorCallResponseWithMessage,
     FieldValidationErrors} from '../utils/call_responses';
-import {tryPromiseWithMessage, isUserSystemAdmin} from '../utils';
+import {tryPromiseWithMessage} from '../utils';
 import {newMMClient, newZDClient} from '../clients';
 import {ZDClientOptions} from 'clients/zendesk';
 import {MMClientOptions} from 'clients/mattermost';
@@ -107,6 +107,11 @@ class AppImpl implements App {
         let request: any;
         let action: string;
         let actionType: string;
+
+        const uniqueSubnameError = {
+            [SubscriptionFields.SubTextName]: 'Channel subscription names must be unique. Please choose another name.',
+        };
+
         const subName = this.values[SubscriptionFields.SubTextName];
         switch (true) {
         case (this.values && this.values[SubscriptionFields.SubmitButtonsName] === SubscriptionFields.DeleteButtonLabel):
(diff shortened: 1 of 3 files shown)


src/app/app.ts
```diff
@@ -150,6 +161,27 @@ class AppImpl implements App {
         return newOKCallResponseWithMarkdown(msg);
     }
 
+    validateSubNameIsUnique = (subName: string): boolean => {
+        const state = this.call.state;
```
mickmister, line 165: What's the reasoning for using `state`? Is it so we don't re-compute the subscriptions across form interactions?

src/app/app.ts
```diff
@@ -150,6 +161,27 @@ class AppImpl implements App {
...
+
+        const subOptions = state.filter((option: AppSelectOption) => option.label === subName);
+        const subSelectLabel = values?.[SubscriptionFields.SubSelectName].label;
+
+        // if changing the subName of an existing subscription, ensure new name does not exist
+        if (subSelectLabel !== subName) {
+            return subOptions.length === 0;
+        }
+
+        return subOptions.length <= 1;
```
mickmister, line 182: Just to be clear, the first if statement here runs if we _are_ changing the name, and we correctly fail when the name is already taken by a different subscription. The final return statement checks that only the subscription we are editing has that name (though we should never be in a state where that's the case).

mickmister (approved):
LGTM, just a few questions. Good work @jfrerich :+1:

src/app/app.ts
jfrerich: Yes, I wanted to minimize the number of Zendesk calls.  I've got the subs available after the first opening of the modal so I just stored it in state and pass it around.

src/app/app.ts
jfrerich: I changed some variable naming to try and help clear this function up.  

You are correct in how this function operates with one exception.  If someone did use this App (probably just Dylan and me, but you never know) the final return function can occur.  Suppose I created 5 subs before this code is added, all with the same name.  The last return will ensure the user changes the name of the sub to something unique.

jfrerich:
Removed QA tag.  Dylan will test at release time.

DHaussermann:
Tested subscriptions change PR
- Subscriptions names are now unique per channel
- If duplicate names where created with the previous version, updating a subscription enforces the validation
- Confirmed duplicate name check is only enforced with a channel as expected and not globally 
- Name duplication check is not case agnostic but this seems consistent with enforcement on the Zendesk side
LGTM!
Removed QA review label.

catalintomai (approved):

PR #88, opened 2021-11-25
[MM-40214] Update icon for App Bar

mickmister:
#### Summary

This PR updates the icon in the manifest, for use int the App Bar.

I also added commands `npm run dev` and `npm run dev:watch` for local development.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-40214


mickmister:
@DHaussermann you can run this locally by pulling and running:

```sh
npm i
npm run build
npm run dev
```

jfrerich (approved):
Thanks, @mickmister!  Could we have used one of the other .png files in `/assets`?

mickmister:
@jfrerich I see there are 4 icons in that folder:

- zendesk-profile-icon-small.png
- zendesk-profile-icon.png
- zendesk.png
- zendesk.svg

Can you describe how each of these are used at the moment? In order to match all themes, we will be placing a white circle behind the icon (possibly in all cases for each icon, though I'm trying to determine that here). cc @matthewbirtch

jfrerich:
* `zendesk-profile-icon.png` used by the manifest and will show post icon for the bot
* `zendesk.png` - used do define `icon` field in forms.  This icon is slightly larger so the icon in the form header isn't so small


* `zendesk.svg` - appears to not be used and can be removed
* `zendesk-profile-icon-small.png` - this looks like it isn't called anywhere in code and can be removed

mickmister:
@jfrerich One is used for the channel header right? I think the one I updated here is that one, the `zendesk.png` file. The channel header binding icon is used in the app bar now at the moment

matthewbirtch:
How do I test this so I can see how it's being used and rendered?

mickmister:
@jfrerich I'm thinking the icon file names should be more descriptive, e.g. `marketplace-and-profile-icon.png`, `channel-header-icon.png`, and potentially `app-bar-icon.png` if we separate that into its own abstraction

PR #90, opened 2021-12-03
[MM-40475] Update to latest version of apps plugin

hanzei:
#### Summary
Targets: https://github.com/mattermost/mattermost-plugin-apps/pull/298

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40475

PR #91, opened 2021-12-03
[MM-40467] Fix panic if oauth2user is empty

hanzei:
#### Summary
If the oauth2 user is not set, the app should not try to access the data. I'm proposing a defensive approach here, which every time checks if the data is there.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40467

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -164,7 +164,7 @@ export function isUserSystemAdmin(actingUser: UserProfile): boolean {
 }
 
 export function isConnected(oauth2user: StoredOauthUserToken): boolean {
-    if (oauth2user && oauth2user.token.access_token && oauth2user.token.access_token !== '') {
+    if (oauth2user?.token?.access_token !== '') {
         return true;
     }
     return false;
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -164,7 +164,7 @@ export function isUserSystemAdmin(actingUser: UserProfile): boolean {
 }
 
 export function isConnected(oauth2user: StoredOauthUserToken): boolean {
-    if (oauth2user?.token?.access_token !== '') {
+    if (oauth2user?.token?.access_token) {
         return true;
     }
     return false;
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -38,7 +38,7 @@ class AppImpl implements App {
         this.context = call.context as CtxExpandedBotActingUserOauth2UserChannelPost;
         this.values = call.values as AppCallValues;
         this.zdOptions = {
-            oauth2UserAccessToken: this.context.oauth2.user.token.access_token,
+            oauth2UserAccessToken: this.context.oauth2.user?.token?.access_token,
             botAccessToken: this.context.bot_access_token,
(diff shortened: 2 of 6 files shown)


jfrerich (approved):

src/utils/utils.ts
```diff
@@ -164,7 +164,7 @@ export function isUserSystemAdmin(actingUser: UserProfile): boolean {
 }
 
 export function isConnected(oauth2user: StoredOauthUserToken): boolean {
-    if (oauth2user && oauth2user.token.access_token && oauth2user.token.access_token !== '') {
+    if (oauth2user?.token?.access_token !== '') {
```
mickmister, line 167: In this case `access_token` will be `undefined`, and we will incorrectly assumed it's defined when we assert that it's not the empty string. An empty string is [treated as falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) so the check below will work:

```suggestion
    if (oauth2user?.token?.access_token) {
```

@jfrerich Any reason for the explicit empty string comparison in the existing code?

mickmister (changes requested):
Just one request to do a `falsy` comparison rather than an explicit empty string comparison on one of the checks.

Also, should the code following the spots modified in this PR be changed to react to the fact that the value may be undefined?

mickmister:
Currently blocking on my comment:

> Also, should the code following the spots modified in this PR be changed to react to the fact that the value may be undefined?

The current change in the PR may result in panics further down in the functions that were edited, or functions that receive these objects containing the edited values. @jfrerich Can you please take a look at the code paths here?

hanzei:
> The current change in the PR may result in panics further down in the functions that were edited, or functions that receive these objects containing the edited values. @jfrerich Can you please take a look at the code paths here?

@mickmister I don' think there is an issue here. The access token is only being used in `newZDClient`, which checks if it's set: https://github.com/mattermost/mattermost-app-zendesk/blob/042d0e50d0f60949e28eda0e9030ca4632676f65/src/clients/zendesk.ts#L67-L69

Calls to `newZDClient` are mostly guarded with a `catch`. The only one missing a catch is `fCreateTarget`. Should we add that?

mickmister:
@hanzei Thanks for going through the code more thoroughly to check for other issues :+1:

> The only one missing a catch is `fCreateTarget`. Should we add that?

If the improvement is easy to add, sure why not add it here. The original problem here is only with the OAuth complete call (I think) so it's not necessary to introduce the error catch in this PR, but it doesn't hurt to add more defensive code when we see the opportunity.

hanzei:
@mickmister Added a catch around it. Maybe it's worth opening a (HW?) ticket to add a `catch` for all `CallResponseHandler`'s?

hanzei:
@mickmister Gentle reminder to review this PR

mickmister (approved):

hanzei:
@DHaussermann Heads up that I'm going to merge this PR as it is as it's it only adds more defensive code and error checks and blocks the review on https://github.com/mattermost/mattermost-app-zendesk/pull/90.

PR #93, opened 2022-01-17
[GH-92] Simplify error handling for error call responses

### LINKED ISSUE #92: Simplify error handling for error call responses
There are lots several 7-line blocks in [src/restapi/fConnect.ts](https://github.com/mattermost/mattermost-app-zendesk/blob/3b94866d2f8e1e0450515d47f3817db55c7b4785/src/restapi/fConnect.ts) for handling errors:

https://github.com/mattermost/mattermost-app-zendesk/blob/3b94866d2f8e1e0450515d47f3817db55c7b4785/src/restapi/fConnect.ts#L114-L120

This could be made into a one-liner by using a helper function:

```ts
await tryCallResponseWithMessage(ppClient.storeOauth2User(storedToken), 'fOauth2Complete - Unable to store oauth2user', res);
```

which could be implemented as

```ts
export async function tryCallResponseWithMessage(promise, message, res) {
    return promise.catch((error) => {
        const fullMessage = message + ': ' + error.message;
        const callResponse = newErrorCallResponseWithMessage(fullMessage);
        res.json(callResponse);
    });
}
```

This improvement would make the code a lot easier to read. This can probably be applied to many places of the application.

bdoodo:
#### Summary
Simplified error handling logic by replacing their try/catch blocks with a helper function stored in utils.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-app-zendesk/issues/92


mickmister (changes requested):
Hi @bdoodo! Welcome to the Mattermost community :tada:

Thanks for this pull request! The code is nice and flat now :+1:

I have one request to apply this strategy in a few other places. Looking through the code, a similar improvement can be added to these files:

- `fCreateTicket.ts`
- `fDisconnect.ts`
- `fSubscriptions.ts`
- `fTarget.ts`

Let me know what you think about this @bdoodo

`+20 -52` Nice diff, I love seeing that!

bdoodo:
Thanks for the encouragement, I'm on it! @mickmister

jfrerich (approved):
Thanks, @bdoodo!  This is fantastic :)

src/utils/utils.ts
```diff
@@ -181,3 +183,11 @@ export function isZdAdmin(role: ZDRole): boolean {
...
     return Boolean(config.zd_target_id && config.zd_target_id !== '');
 }
+
+export async function tryCallResponseWithMessage(promise, message, res) {
+    return promise.catch((error) => {
+        const fullMessage = message + ': ' + error.message;
+        const callResponse = newErrorCallResponseWithMessage(fullMessage);
+        res.json(callResponse);
+    });
+}
```
mickmister, line 193: nit: Please add a blank newline at the end of this function

src/restapi/fTarget.ts
```diff
@@ -17,15 +17,14 @@ export const fCreateTarget: CallResponseHandler = async (req, res) => {
...
-        res.json(callResponse);
-    }
+    await tryCallResponseWithMessage(
+        updateOrCreateTarget(zdClient, context).then((text) => {
+            const callResponse = newOKCallResponseWithMarkdown(text);
+            res.json(callResponse);
+        }),
+        'Unable to create target',
+        res
+    );
```
mickmister, line 27: What do you think about avoiding the usage of `then()` when calling 
`tryCallResponseWithMessage`? I think this looks cleaner:

```suggestion
    const text = await tryCallResponseWithMessage(
        updateOrCreateTarget(zdClient, context),
        'Unable to create target',
        res
    );

    const callResponse = newOKCallResponseWithMarkdown(text);
    res.json(callResponse);
```

I do like the multiple-line approach here though. It's clear that there is special error handling going on this way.

mickmister (changes requested):
Thanks for adding the improvements @bdoodo! I have a request to avoid using `then()` in the spots we've refactored here. The main purpose of these improvements is to make the code more readable, and more specifically more "flat".

When we have indented code, it's difficult for a developer to parse the important parts when reading the code. If we treat each `await`'d statement as a discrete operation, with no important logic indented, then it's much easier to read.

I apologize for taking so long to get back to this. To make sure the reviewers are aware that they need to re-review the PR, please use GitHub's review request feature, by clicking the icon with arrows here:

image

src/restapi/fTarget.ts
bdoodo: That's much cleaner. I'll go through and make similar changes. Thank you for teaching me about codebase conventions as well.

bdoodo:
Sorry for the long delay; my courseload has been heavier than anticipated. Moving forward in future tasks, I'll make sure to communicate timing difficulties and only take on tasks I'm very confident I'll be able to commit to for their full review cycle.

src/restapi/fConnect.ts
```diff
@@ -111,21 +81,9 @@ export const fOauth2Complete: CallResponseHandler = async (req, res) => {
...
-    try {
-        await app.createBotDMPost(dmText);
-    } catch (error) {
-        callResponse = newErrorCallResponseWithMessage('fOauth2Complete - Unable to create bot DM post: ' + error.message);
-        res.json(callResponse);
-        return;
-    }
-    callResponse = newOKCallResponse();
+    await tryCallResponseWithMessage(app.createBotDMPost(dmText), 'fOauth2Complete - Unable to create bot DM post', res);
+    const callResponse: AppCallResponse = newOKCallResponse();
```
mickmister, line 87: I genuinely feel that you can ignore this request, as there has already been a decent amount back-and-forth on the PR, and I don't want to bog down you with more requests. Though here is the suggestion:

I suggest we add a newline after every `tryCallResponseWithMessage` operation, mainly to improve readability.

```suggestion
    await tryCallResponseWithMessage(app.createBotDMPost(dmText), 'fOauth2Complete - Unable to create bot DM post', res);

    const callResponse: AppCallResponse = newOKCallResponse();
```

mickmister (approved):
LGTM! Just one remaining suggestion to add a newline after each of the `tryCallResponseWithMessage` operations

src/restapi/fConnect.ts
bdoodo: Thanks Mick, I'm enjoying learning about codebase conventions. So since the emphasis would be on making important code readable, I should also add a space above if it looks cramped, and it's more of an art than a science, right? Feels like a silly question to ask, but when the scope is on spacing, I want to make sure I'm not making a mistake haha

mickmister: > So since the emphasis would be on making important code readable, I should also add a space above if it looks cramped

The convention is to "place newlines _after_ blocks". If it is a block that is preceeding the call to `tryCallResponseWithMessage` then there should be a newline. In the case of this project, a block would be any multi-line statement, whether that be a function call with its arguments are expanded across multiple lines, or an actual block such as an `if` block or `for` loop block.

hanzei:
@bdoodo Would you please merge merge `master` into your PR?

PR #95, opened 2022-04-20
Support zendesk webhooks instead of targets

mickmister:
#### Summary

Zendesk targets are no longer supported by Zendesk. This PR makes it so we use the Webhooks API instead of the Targets API to create subscriptions. Triggers work the same way as they did with targets, so that logic doesn't change on our side.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-42566


jfrerich (approved):
LGTM!! nice work!

mickmister:
@DHaussermann @levb This PR includes changes from https://github.com/mattermost/mattermost-app-zendesk/pull/94 to make this ready for a new release with Apps 1.0

mickmister (commented):
small improvement

mickmister:
@DHaussermann I've updated the zendesk project to reflect the changes in framework's expand logic. All features seem to be working properly (create ticket, webhooks)

DHaussermann:
@mickmister I have done another round of testing on this. 

The issue with the `connect` binding not appearing has been resolved. However I am not able to connect after the application has been configured. 

When I click the connection link - I reach service now and it find my oAuth app on the service now side. However, after I click "allow" I see an error that the app can't be found. 
image
image
image

Maybe something is wrong with the App ID it's looking for?
logs show:
`{"timestamp":"2022-08-12 10:32:53.857 -04:00","level":"warn","msg":"Failed to complete remote OAuth2","caller":"app/plugin_api.go:978","plugin_id":"com.mattermost.apps","from_user_id":"zrro3i41biyh9qtm3jnqw3g96o","source_plugin_id":"<nil>","destination":"com.mattermost.zendesk","path":"/apps/com.mattermost.zendesk/oauth2/remote/complete","request_id":"n6dtqstyhtbpbm65fzyifurqwc","error":"app com.mattermost.zendesk is not installed: not found"}`

Thoughts?

mickmister:
@DHaussermann The id of the app has changed from `com.mattermost.zendesk` to `zendesk`. I think your OAuth redirect URL set in the Zendesk UI needs to be changed to use the new id

DHaussermann:
@mickmister your suggestion for updating the oAuth URL to match the current app ID did resolve the issue with authentication.

I was able to connect a user and setup subscriptions. 

That said I was not able to see subscription events delivering in Mattermost when they matched the criteria as expected.

DHaussermann:
@mickmister another follow-up here. 
I deleted all the relevant data from the test instance and started a fresh setup. 

What I found is that it **seems** like the webhook get's created on the ZD side with no secret set.
image

Thoughts?

mickmister:
@DHaussermann The undefined secret issue should be fixed now. Please take another look when you have the chance :+1:

DHaussermann:
@mickmister can we connect on this?

I have the latest deployed but I still see undefined when configuring a webhook.
image.

mickmister:
@DHaussermann You may need to run `/zendesk setup-webhook` again if you haven't run it. I'm successfully setting the URL with a secret on my setup.

Here's what the webhooks looked like before I cleaned them up just now:

image

Notice the quantity and duplication among the entries. I'm thinking we should make the Zendesk app be aware of existing webhooks, and only create them if necessary. And possibly support rotating of webhook secret.

mickmister:
@DHaussermann I'm seeing an issue where the subscribe modal is not showing the fields when I select "Create New". Looks like this may be a bug in the webapp since the form is being returned to the client with the fields included:

image

mickmister:
Made a ticket for the above issue https://mattermost.atlassian.net/browse/MM-49185

DHaussermann:
@mickmister not sue if there are any issues with subscriptions themselves. 

I was able to make one and see it delivering.

But I see this issue where the app crashes
image

Maybe at this point, this and the related PR should be merged and we can investigate this separately? 
Not sure if we're ready to begin testing this deployed as a lambda until this is sorted?

mickmister:
@DHaussermann It's strange that the subscription post was created, yet this error occurred here. Doesn't make sense to me. Thinking about it some more, I think we should merge without fixing that issue, for the purpose of moving this release along. Let's sync up tomorrow on what may be going on here.

